PR #8, opened 2017-07-14
SDK-20 : Add getter for user profile attributes

sidibos:


tests/ActivityDetailsTest.php
```diff
@@ -0,0 +1,142 @@
+<?php
+
+use \Yoti\ActivityDetails;
+
+define('DUMMY_SELFIE_FILE', __DIR__ . '/../src/sample-data/dummy-avatar.png');
```
vassyz: Are you sure this image is not copyrighted?

tests/ActivityDetailsTest.php
```diff
@@ -0,0 +1,142 @@
...
+     */
+    public function testGetSelfie()
+    {
+        $this->assertEquals($this->dummyProfile['selfie'], $this->profile->getSelfie());
+    }
+
+    /**
+     * Test we are not returning a wrong given name
+     */
+    public function testWrongGivenNames()
```
vassyz: I think we either add wrong assertions for all of the setters or for none.

sidibos: I think there is no need to add wrong assertion for all the getters, in this case I think it's better not have it otherwise the test case will get bigger.

tests/ActivityDetailsTest.php
sidibos: I have added a free image which shouldn't have any copyright issue.

src/Yoti/ActivityDetails.php
```diff
@@ -113,4 +115,152 @@ public function getUserId()
...
+    /**
+     * Handles a call to an undefined function
+     * @param $method
+     * @param $args
+     * @return mixed|null|void
+     */
+    public function __call($method, $args)
+    {
+        // Check method name is alphanumeric
+        if (!preg_match('~^[0-9a-z]+$~i', $method)) {
```
vassyz: Not sure what the conventions are, but I'm pretty sure you should allow underscores as well.

README.md
```diff
@@ -91,6 +91,49 @@ else
 }
 ``` 
 
+### Available User profile attributes through getters
+We have exposed user profile attributes through getters. You will find in the example below the getters available to you and how to use them:
```
vassyz: exposed *the* user profile

vassyz: Do you agree with this one?

sidibos: Not really, it's not specific to a user as it's for any user. But if you want to add it there I will add it.

vassyz:
Looks good, apart from that missing _the_. Could you also squash the commits it they can't be separated by logic? I think I mentioned this in the last PR. It just makes the git history cleaner.

README.md
vassyz: It's fine, leave it as it is.

PR #10, opened 2017-10-17
[SDK-130]: Changed license to MIT

echarrod:
I found another license in src/LICENSE. This is an Apache one.

It doesn't look like it relates to anything - I created it in a separate commit in case I'm missing something obvious and we do actually need it, let me know.

There are 3 LICENSE files in directories below this, but they're all using MIT.


echarrod:
I've added the license to the /src folder too now in the last commit

echarrod:
After discussion, since the plugin will have it's own license, we're not committing the MIT license into /src too

PR #11, opened 2017-10-18
[SDK-130]: Added license to composer.json file

echarrod:
Just creating a PR to double check this is right


echarrod:
No worries!

PR #12, opened 2017-10-31
SDK-117 : Add X-Yoti-SDK header to the request and remove Fullname attribute

sidibos:


src/Yoti/YotiClient.php
```diff
@@ -8,24 +8,32 @@
...
     const CONNECT_BASE_URL = 'https://www.yoti.com/connect';
 
-    // dashboard login
+    // Dashboard login
     const DASHBOARD_URL = 'https://www.yoti.com/dashboard';
 
+    // Accepted HTTP header values for X-Yoti-SDK header
+    const YOTI_ACCEPTED_SDK_HEADERS = [
+        'PHP',
+        'Wordpress',
```
vassyz: Shouldn't it be `WordPress`

src/Yoti/YotiClient.php
```diff
@@ -52,50 +60,63 @@ class YotiClient
     private $_mockRequests = false;
 
     /**
+     * @var string
+     */
+    private $_yotiSdkHeader;
```
vassyz: Other SDKs are using the name `sdkIdentifier`, up to you if you want to use the same.

sidibos: yes I will use the sdkIdentifier instead to keep consistency.

src/Yoti/YotiClient.php
sidibos: Ok I will use `WordPress`

PR #14, opened 2018-01-08
SDK-166 : Exclude example and tests from published packages.

sidibos:


example/profile.php
```diff
@@ -1,8 +1,11 @@
 <?php
-require_once './vendor/autoload.php'; // Make sure you run composer update inside the example folder before trying them
+
+// Make sure you run composer update inside the example folder before trying them
```
vassyz: If it's just an example it should read: `example folder before trying it out`

example/profile.php
```diff
@@ -1,8 +1,11 @@
...
+require_once './vendor/autoload.php';
+// Error message
```
vassyz: What purpose does this comment serve?

example/profile.php
```diff
@@ -1,8 +1,11 @@
...
 $errorMsg = '';
 // Get and sanitize the token
-$token = (isset($_GET['token']))? preg_replace('/[^-a-zA-Z0-9_]/', '', $_GET['token']) : '';
+$token = isset($_GET['token']) ? preg_replace('/[^-a-zA-Z0-9_]/', '', $_GET['token']) : '';
```
vassyz: Do we need this step? `preg_replace('/[^-a-zA-Z0-9_]/'`

example/profile.php
sidibos: Yes this because the example needs all the dependencies to work, which will be generated in the vendor folder when you run composer update.

example/profile.php
```diff
@@ -1,8 +1,11 @@
...
 $config = [
     'sdkId' => 'add your SDK ID here', // This is your SDK ID associated with the Yoti Application you created on Dashboard
     'pemFile' => __DIR__ . '/keys/your-key-name.pem', // This is the public key (in .pem format) associated with the Yoti Application you created on Dashboard
```
vassyz: I think this is the `private` key, not public.

example/profile.php
sidibos: I think we need it, it's sanitising the token, it's removing special characters from the token.

example/profile.php
sidibos: Yes you are right, I will amend that.

example/profile.php
vassyz: Shouldn't the SDK just decode what it receives from Connect?

sidibos: Yes it should indeed, but we are trying to prevent any injection in case someone tries to modify the token because it's not a Post but a Get. We are just trying to show people how to make their code robust.
I am happy to remove it you still think it's not needed.

PR #15, opened 2018-01-15
SDK-138 : Add base64Selfie helper method

sidibos:


example/profile.php
```diff
@@ -28,16 +29,16 @@
...
-            <strong>Nationality</strong> <?php echo $profile->getNationality(); ?><br>
+            <strong>Given Name(s)</strong> <?php echo $profile->getGivenNames() ?><br>
+            <strong>Family Name</strong> <?php echo $profile->getFamilyName() ?><br>
+            <strong>Phone</strong> <?php echo $profile->getPhoneNumber() ?><br>
+            <strong>Email</strong> <?php echo $profile->getEmailAddress() ?><br>
+            <strong>Date Of Birth</strong> <?php echo $profile->getDateOfBirth() ?><br>
+            <strong>Address</strong> <?php echo $profile->getPostalAddress() ?><br>
+            <strong>Gender</strong> <?php echo $profile->getGender() ?><br>
+            <strong>Nationality</strong> <?php echo $profile->getNationality() ?><br>
             <strong>Photo</strong><br> <img src="data:image/x-icon;base64,<?php echo $selfie ?>" />
```
vassyz: Shouldn't you be using `getBase64Selfie` here?

sidibos: Yes I will add that in the next task SDK-142 - I din't want it in this PR otherwise the example will break if someone try to use it by running composer update as the version including the helper won't be released.

PR #16, opened 2018-01-17
SDK-142 : Use ActivyDetails helper in the example and amend user profile display

sidibos:


example/index.html
```diff
@@ -1,20 +1,24 @@
...
     </head>
     <body>
         <h1>Your Site Information</h1>
         <span data-yoti-application-id="your App ID"
             data-yoti-type="inline"
-            data-yoti-scenario-id="your Scenario ID"
-            data-size="small">
-         Use Yoti!
+            data-yoti-scenario-id="your Scenario ID">
+         USE YOTI!
```
vassyz: Please don't use unnecessary uppercase and exclamation marks `Use Yoti` on the button will do.

example/profile.php
```diff
@@ -25,22 +35,46 @@
 <html>
    <head>
         <meta charset="utf-8">
-        <title>PROFILE PAGE</title>
+        <title>YOTI PROFILE</title>
```
vassyz: No uppercase

example/profile.php
```diff
@@ -3,8 +3,12 @@
 // Make sure you run composer update inside the example folder before trying this example out
 require_once './vendor/autoload.php';
 
+use Yoti\Helper\ActivityDetailsHelper;
+
 // Log any error message
 $errorMsg = '';
+// Selfie file name.
+$selfieFile = 'selfie.jpeg';
```
vassyz: I would remove this.

example/profile.php
```diff
@@ -16,7 +20,13 @@
...
     $yotiClient = new Yoti\YotiClient($config['sdkId'], $config['pemFile']);
     $profile = $yotiClient->getActivityDetails($token);
-    $selfie = base64_encode($profile->getSelfie());
+    $base64Selfie = ActivityDetailsHelper::getBase64Selfie($profile);
+    $selfieEntity = $profile->getSelfieEntity();
+    // Retrieve selfie image format
+    $imageFormat = $selfieEntity ? $selfieEntity->getType() : 'jpeg';
+    $selfieFile = "selfie.{$imageFormat}";
+    // Create selfie image file.
+    file_put_contents($selfieFile, $profile->getSelfie(), LOCK_EX);
```
vassyz: I understand the logic, but I would minimise the code here for simplicity's sake.

```
$base64Selfie = ActivityDetailsHelper::getBase64Selfie($profile);
file_put_contents('selfie.jpeg', $profile->getSelfie(), LOCK_EX);
```

vassyz: I would also use a `uploads` or `public` folder to save the selfie. The content in this folder should be ignored by git, but also add a `.keep` file inside.

example/profile.php
```diff
@@ -25,22 +35,46 @@
...
+                <dd><?php echo $profile->getPostalAddress() ?></dd>
+
+                <dt>Gender</dt>
+                <dd><?php echo $profile->getGender() ?></dd>
+
+                <dt>Nationality</dt>
+                <dd><?php echo $profile->getNationality() ?></dd>
+
+                <dt>Selfie as base64 data</dt>
+                <dd><img src="<?php echo $base64Selfie ?>" /></dd>
```
vassyz: I feel like the `base64Selfie` should be inside the `$profile` object, but I can also understand why you want to have it in a separate helper.

example/index.html
sidibos: I used uppercase to be inline with the QR code button as it will display the text in uppercase anyway.

example/profile.php
sidibos: I wouldn't mind hard coding the image file name, but I was trying to show to the user how they could get the selfie image file format.

example/profile.php
sidibos: That's fine I can use snake-case format.

PR #21, opened 2018-02-27
SDK-270 : Add tests to PHP AML check

sidibos:


example/.env.profile
```diff
@@ -0,0 +1,8 @@
+# This file is a template for defining the environment variables
+# Set the application config values here
+
+YOTI_APP_ID=388f7483-080e-48c5-b2fb-668783767d2c
```
vassyz: Don't put actual values here. I prefer naming these files `.env.example`, but it's up to you.

src/Yoti/YotiClient.php
```diff
@@ -195,35 +195,44 @@ public function getActivityDetails($encryptedConnectToken = NULL)
...
+                $signedRequest->getApiRequestUrl($this->_connectApi),
+                $amlPayload,
+                RestRequest::METHOD_POST
+            );
 
-        $result = $restRequest->exec();
+            $result = $restRequest->exec();
+        }
+        else {
+            // Sample AML result, don't make curl call instead spoof response from aml-check-result.json
```
vassyz: Is it not possible to only have this in the tests?

src/sample-data/aml-check-private-key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
vassyz: Shouldn't these files be in the tests folder?

tests/YotiClientTest.php
```diff
@@ -1,11 +1,15 @@
...
+use Yoti\Entity\Country;
+use Yoti\Entity\AmlAddress;
+use Yoti\Entity\AmlProfile;
+use Yoti\Http\AmlResult;
 
-define('PEM_FILE', __DIR__ . '/../src/sample-data/yw-access-security.pem');
-define('SDK_ID', '990a3996-5762-4e8a-aa64-cb406fdb0e68');
-define('YOTI_CONNECT_TOKEN', file_get_contents(__DIR__ . '/../src/sample-data/connect-token.txt'));
-define('INVALID_YOTI_CONNECT_TOKEN', 'sdfsdfsdasdajsopifajsd=');
+defined('PEM_FILE') || define('PEM_FILE', __DIR__ . '/../src/sample-data/yw-access-security.pem');
```
vassyz: How can this be defined already? And if it is, shouldn't we re-define it before running the following tests?

example/.env.profile
sidibos: That file shouldn't have been pushed - it got mixed up with the commits - I will remove it.

src/sample-data/aml-check-private-key.pem
sidibos: Most tests files are inside the `src/sample-data` - this file can be moved inside the `tests` folder and will have to move all files that are only used in `tests` only.

src/Yoti/YotiClient.php
sidibos: Yes anything is possible - I will move that block of codes inside a function - that way the function can be mocked during the test so no call will made to connect API.

tests/YotiClientTest.php
sidibos: The same constant is defined in `RestRequestTest` with the same value - because UNIT TEST is loading all the files in no particular order when you run the test - then you will end up with an errors if you don't add that condition.

example/scripts/aml-check-uk.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
vassyz: Please name this file `aml-check.php`. It's not UK-specific.

vassyz:
The Readme has `How to run the example - How to run the example`. You can remove what's after `-`.

examples/scripts/aml-check.php
```diff
@@ -0,0 +1,32 @@
...
+// Load dependent packages and env data
+require_once __DIR__ . '/../bootstrap.php';
+
+use Yoti\Entity\Country;
+use Yoti\Entity\AmlAddress;
+use Yoti\Entity\AmlProfile;
+use Yoti\YotiClient;
+
+//$connectAPI = YotiClient::DEFAULT_CONNECT_API;
+$connectAPI = 'https://staging0.api.yoti.com:8443/api/v1';
```
vassyz: This should probably not be here.

sidibos: Oh shoot forgot it there - thx

PR #22, opened 2018-03-06
SDK-275 : Add age verification helper

sidibos:


examples/profile.php
```diff
@@ -17,7 +17,9 @@
     $base64Selfie = ActivityDetailsHelper::getBase64Selfie($profile);
     // This key uses the  format: age_[over|under]:[1-999] and is dynamically
```
vassyz: Remove this comment

examples/profile.php
```diff
@@ -17,7 +17,9 @@
...
     // generated based on the dashboard attribute Age / Verify Condition
-    $ageVerified = $profile->getProfileAttribute('age_over:18');
+    $ageVerified = $profile->isAgeVerified() ? 'true' : '';
+    $verifiedAge = $profile->getVerifiedAge();
```
vassyz: Maybe add a comment of what this does?

examples/profile.php
```diff
@@ -17,7 +17,9 @@
     $base64Selfie = ActivityDetailsHelper::getBase64Selfie($profile);
     // This key uses the  format: age_[over|under]:[1-999] and is dynamically
     // generated based on the dashboard attribute Age / Verify Condition
-    $ageVerified = $profile->getProfileAttribute('age_over:18');
+    $ageVerified = $profile->isAgeVerified() ? 'true' : '';
```
vassyz: What does this do? Are you converting it to a string or not wanting to display `false`?

tests/Util/Age/AgeUnderOverProcessorTest.php
```diff
@@ -0,0 +1,60 @@
...
+    public $processor;
+    public $dummyProfile;
+
+    public function setUp()
+    {
+        $this->dummyProfile = [
+            'family_name' => 'TestFamilyName',
+            'given_names' => 'TestGivenName',
+            'full_name'   => 'TestGivenName TestFamilyName',
+            'date_of_birth' => '11-07-2017',
```
vassyz: I know it doesn't matter, but can you add a date so the fake profile is over 18?

tests/Util/Age/ProcessorTest.php
```diff
@@ -0,0 +1,46 @@
...
+     */
+    public $processor;
+
+    public function setUp()
+    {
+        $dummyProfile = [
+            'family_name' => 'TestFamilyName',
+            'given_names' => 'TestGivenName',
+            'full_name'   => 'TestGivenName TestFamilyName',
+            'date_of_birth' => '11-07-2017',
```
vassyz: Same here, age > 18

examples/profile.php
```diff
@@ -17,7 +17,14 @@
     $base64Selfie = ActivityDetailsHelper::getBase64Selfie($profile);
     // This key uses the  format: age_[over|under]:[1-999] and is dynamically
     // generated based on the dashboard attribute Age / Verify Condition
-    $ageVerified = $profile->getProfileAttribute('age_over:18');
+    $ageVerified = $profile->isAgeVerified();
+    if(is_bool($ageVerified)) {
```
vassyz: This conditional doesn't look alright to me. Let's look into it a bit more.

sidibos: This function will return two types a boolean `TRUE or FALSE` if the attribute is set or a string `NULL` if the attribute is not set.

PR #24, opened 2018-03-28
SDK-301 : Add "structured_postal_address" getter

sidibos:


src/Yoti/ActivityDetails.php
```diff
@@ -18,12 +18,14 @@ class ActivityDetails
     const ATTR_GIVEN_NAMES = 'given_names';
     const ATTR_FULL_NAME = 'full_name';
     const ATTR_DATE_OF_BIRTH = 'date_of_birth';
+    const ATTR_AGE_VERIFIED = 'age_verified';
```
vassyz: How was this working before?

src/Yoti/ActivityDetails.php
```diff
@@ -18,12 +18,14 @@ class ActivityDetails
...
     const ATTR_GENDER = 'gender';
     const ATTR_NATIONALITY = 'nationality';
     const ATTR_PHONE_NUMBER = 'phone_number';
     const ATTR_SELFIE = 'selfie';
     const ATTR_EMAIL_ADDRESS = 'email_address';
     const ATTR_POSTAL_ADDRESS = 'postal_address';
+    const ATTR_STRUCTURE_POSTAL_ADDRESS = 'structured_postal_address';
```
vassyz: STRUCTURE or STRUCTURED?

src/Yoti/ActivityDetails.php
sidibos: Basically that attribute is not coming from connect, the age derivation value is extracted by a regex. This constant is mainly for the plugins. I won't have to define it for each plugin.

src/Yoti/ActivityDetails.php
sidibos: Will fix that - it should be structured.

PR #27, opened 2018-05-09
SDK-343-profile-anchors: Add profile attributes anchors

sidibos:
I have added a sample of the new Profile to the doc::## API Coverage section. This is so we can agree how it will be added to the doc and I will finish that section later,


tests/Entity/AttributeTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+    public function testAttributeValue()
+    {
+        $this->assertEquals(self::ATTR_VALUE, $this->dummyAttribute->getValue());
+    }
+
+    public function testAttributeSources()
+    {
+        $this->assertEquals(
+            json_encode(self::ATTR_SOURCES),
```
vassyz: Is it necessary to encode twice?

vassyz:
Is it not possible to add `phpseclib` to Composer instead of the SDK source?

I also think the `Running the tests` section should be updated with more details.

tests/Entity/AttributeTest.php
sidibos: We are comparing 2 arrays, so the best way is to compare there JSON strings.

sidibos:
Yes it's possible, but like protobuf we are including `phpsec` directly for applications that cannot use composer - like the plugins for instance. When we start using `composer` for all the plugins we could think about using it in the PHP SDK too. But even though there are some cases where composer is not allowed for security reason and those applications will need to use our SDK without the need of composer.

sidibos:
I could add more details in `running the tests` section, for example the cmd to use.

sidibos:
@qzagarese would you want to have a final look before I merge this PR?

examples/profile.php
```diff
@@ -70,89 +69,412 @@
...
+                            <th scope="col">Name</th>
+                            <th scope="col">Value</th>
+                            <th scope="col">Anchors</th>
+                        </tr>
                     </thead>
                     <tbody>
                         <tr>
                             <th scope="row">Given Name(s)</th>
                             <td><?php echo $givenNames ? $givenNames->getValue() : '' ?></td>
                             <td>
```
echarrod: We also have the Yoti ID/`rememberMeID` for .NET, and I think some of the others, but obviously not showing anything for the anchors, just `N/A`

examples/profile.php
```diff
@@ -70,89 +69,412 @@
...
                         <tr>
                             <th scope="row">Family Name</th>
                             <td><?php echo $familyName ? $familyName->getValue() : '' ?></td>
-                            <td><?php echo ProfileHelper::getAttributeSources($familyName) ?></td>
-                            <td><?php echo ProfileHelper::getAttributeVerifiers($familyName) ?></td>
+                            <td>
+                                <table class="table">
+                                    <thead class="thead-light">
+                                        <tr>
+                                            <th>S/V</th>
```
echarrod: Do you think it would be better to just have these column headers at the top, so we're not having to repeat them each time?

examples/profile.php
```diff
@@ -70,89 +69,412 @@
...
                         <tr>
                             <th scope="row">Full Name</th>
                             <td><?php echo $fullName ? $fullName->getValue() : '' ?></td>
-                            <td><?php echo ProfileHelper::getAttributeSources($fullName) ?></td>
-                            <td><?php echo ProfileHelper::getAttributeVerifiers($fullName) ?></td>
+                            <td>
+                                <table class="table">
+                                    <thead class="thead-light">
+                                        <tr>
+                                            <th>S/V</th>
```
echarrod: @qzagarese I assumed `S/V` was shorthand, and we'd put `Source/Verifier` in the final table, what do you think?

examples/profile.php
sidibos: I don't think we need to display rememberMeId here.

examples/profile.php
sidibos: Maybe, but And I remember Quirino said not have them as a column.

examples/profile.php
sidibos: I think S/V is just the header title the value itself is displaying as 'Source' or 'Verifier'.

src/Yoti/Entity/Anchor.php
```diff
@@ -0,0 +1,108 @@
...
+        $signature,
+        $artifactSignature,
+        $signedTimeStamp,
+        array $originServerCerts
+    ) {
+        $this->value = $value;
+        $this->subType = $subType;
+        $this->signature = $signature;
+        $this->artifactSignature = $artifactSignature;
+        $this->signedTimeStamp = $signedTimeStamp;
```
qzagarese: Here you need to convert this to a SignedTimestamp object. I believe this is a `\Protobuf\Stream` object, so you need to create the class from this proto https://github.com/getyoti/yoti-dotnet-sdk/blob/master/src/Yoti.Auth/ProtoBuf/common-public-api/compubapi_v1/SignedTimestamp.proto
and then create an instance using the fromStream method.

src/Yoti/Entity/Anchor.php
```diff
@@ -0,0 +1,108 @@
...
+        $this->originServerCerts = $originServerCerts;
+
+        $config = new \Protobuf\Configuration();
+        $streamReader = $config->getStreamReader();
+        $this->timeStamp = $streamReader->readFixed32($this->signedTimeStamp);
```
qzagarese: Once you have converted the stream to an object as explained above, you can delete this as that object should have a `getTimestamp` method itself.

src/Yoti/Entity/Anchor.php
```diff
@@ -0,0 +1,108 @@
...
+     * @return \Protobuf\Stream
+     */
+    public function getSignedTimeStamp() {
+        return $this->signedTimeStamp;
+    }
+
+    /**
+     * @return int
+     */
+    public function getTimeStamp() {
```
qzagarese: this can be removed.

src/Yoti/Entity/Anchor.php
```diff
@@ -0,0 +1,108 @@
...
+     * @return \Protobuf\Stream
+     */
+    public function getArtifactSignature() {
+        return $this->artifactSignature;
+    }
+
+    /**
+     * @return \Protobuf\Stream
+     */
+    public function getSignedTimeStamp() {
```
qzagarese: this should now return a `SignedTimestamp` instance.

README.md
```diff
@@ -167,11 +173,29 @@ if ($client->getOutcome() == \Yoti\YotiClient::OUTCOME_SUCCESS) {
...
+You can also retrieve further properties from these respective anchors in the following way:
+
+```php
+<?php
+// Retrieving properties of the first anchor
+$value = $givenNamesSources[0]->getValue(); // string
+$subType = $givenNamesSources[0]->getSubType(); // string
+$signature = $givenNamesSources[0]->getSignature(); // bytes
+$artifactSignature = $givenNamesSources[0]->getArtifactSignature(); // bytes
+$timeStamp = $givenNamesSources[0]->getTimeStamp(); // integer
```
qzagarese: This can be removed. Please see comments below.

README.md
```diff
@@ -167,11 +173,29 @@ if ($client->getOutcome() == \Yoti\YotiClient::OUTCOME_SUCCESS) {
...
+$signedTimeStamp = $givenNamesSources[0]->getSignedTimeStamp(); // bytes
```
qzagarese: this should return a `SignedTimestamp` instance. Below I've explained how to create one.
Then you can add a line like the following: `$timeStamp = $signedTimeStamp->getTimeStamp(); // int `

README.md
```diff
@@ -167,11 +173,29 @@ if ($client->getOutcome() == \Yoti\YotiClient::OUTCOME_SUCCESS) {
...
-The set of attributes the user has configured for the transaction.
+You can also retrieve further properties from these respective anchors in the following way:
+
+```php
+<?php
+// Retrieving properties of the first anchor
+$value = $givenNamesSources[0]->getValue(); // string
+$subType = $givenNamesSources[0]->getSubType(); // string
+$signature = $givenNamesSources[0]->getSignature(); // bytes
+$artifactSignature = $givenNamesSources[0]->getArtifactSignature(); // bytes
```
qzagarese: We can remove this method it as it does not add value to who integrates.

src/Yoti/Entity/Anchor.php
```diff
@@ -0,0 +1,108 @@
...
+     * @var string
+     */
+    protected $subType;
+
+    /**
+     * @var \Protobuf\Stream
+     */
+    protected $signature;
+
+    protected $artifactSignature;
```
qzagarese: this one can be removed.

src/Yoti/Entity/Anchor.php
```diff
@@ -0,0 +1,108 @@
...
+     * @return \Protobuf\Stream
+     */
+    public function getSignature() {
+        return $this->signature;
+    }
+
+    /**
+     * @return \Protobuf\Stream
+     */
+    public function getArtifactSignature() {
```
qzagarese: This one can be removed.

sidibos2:
@qzagarese to answer your question:
Yes it's possible to use composer, but like protobuf we are including phpsec directly because some applications that integrate with Yoti don't use composer - like the plugins for instance. 
When we start using composer for all the plugins we could think about those PHP SDK dependencies into composer. 
But even though there are some cases where composer is not allowed by some companies, for security reason and those applications will need to use our SDK without the need of composer.

README.md
sidibos: Ok I will remove it. I remember seeing this method in the dot-net SDK, maybe it should be removed there as well @echarrod

echarrod: I thought we were just not exposing .GetArtifactLink()? This method was in the Google Doc, and also in the "Planning the next major release" wiki

qzagarese:
@sidibos2 Fair enough :+1:

qzagarese:
LGTM. Nicely done Moussa :+1:

sidibos:
thx!

PR #30, opened 2018-07-09
SDK-377: Self Signed Certificate

sidibos:


README.md
```diff
@@ -258,25 +258,60 @@ The examples can be found in the [examples folder](https://github.com/getyoti/ph
...
 * Create your application in the [Yoti Dashboard](https://www.yoti.com/dashboard/applications) (this requires having a Yoti account)
 * Do the steps below inside the [examples folder](https://github.com/getyoti/php/tree/master/examples)
 * Copy `.env.dist` to `.env` and fill in the environment variables.
+
+Please follow one of the steps below to complete the process:
+    
+#### 1. Running with Docker
+
+* With Docker, the .pem file should reside within the same location where Docker is run from. So make sure you put the .pem file inside the `keys` folder within the [examples folder](https://github.com/getyoti/php/tree/master/examples).
+* Update the YOTI_KEY_FILE_PATH environment variable in the`.env` file as follow `./keys/your-pem-file-name.pem`
```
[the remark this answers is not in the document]
sidibos: Yep it makes sense.

echarrod:
Can we add the .pem file which is placed inside the /keys folder for docker to be ignored by Git? First of all to encourage good practice of not checking in .pem files to source control, and secondly so it doesn't appear in the pending changes

README.md
```diff
@@ -258,43 +258,23 @@ The examples can be found in the [examples folder](https://github.com/getyoti/ph
 ### Profile sharing
```
echarrod: Do you think it would be useful to have a section something like this here so that the users will know that they will need Docker before they get to the penultimate step:


#### Requirements
* [Docker](https://www.docker.com/community-edition)

sidibos: I don't think we need to specify that, but will just add in the title.

bucky-boy: Would be better if we linked to the shortest possible instructions for installing Docker, but maybe later

PR #32, opened 2018-09-07
SDK-540: Add the ability to retrieve DocumentDetails as an Attribute

sidibos:


src/Yoti/ActivityDetails.php
```diff
@@ -147,6 +138,26 @@ public static function constructFromAttributeList(AttributeList $attributeList,
         return $inst;
     }
 
+    private static function getYotiAttribute(ProtobufAttribute $protobuAttribute, array $attributeAnchors, $attrName)
```
bucky-boy: createYotiAttribute?

src/Yoti/ActivityDetails.php
```diff
@@ -103,29 +106,17 @@ public static function constructFromAttributeList(AttributeList $attributeList,
...
-            $attribute = new Attribute(
-                $attrName,
-                $attrValue,
-                $attributeAnchors['sources'],
-                $attributeAnchors['verifiers']
-            );
-
-            $profileAttributes[$attrName] = $attribute;
+            $yotiAttribute = self::getYotiAttribute($item, $attributeAnchors, $attrName);
+            $profileAttributes[$attrName] = $yotiAttribute;
```
bucky-boy: Should all this be inside a null check on _$yotiAttribute_?  I don't see any point adding a _null_ into the _$profileAttributes_ map.

tests/Entity/DocumentDetailsTest.php
```diff
@@ -0,0 +1,78 @@
...
+    public function testShouldParseTwoOptionalAttributes()
+    {
+        $document = new DocumentDetails('DRIVING_LICENCE GBR 1234abc 2016-05-01 DVLA');
+        $this->assertEquals('DRIVING_LICENCE', $document->getType());
+        $this->assertEquals('GBR', $document->getIssuingCountry());
+        $this->assertEquals('1234abc', $document->getDocumentNumber());
+        $this->assertEquals('2016-05-01', $document->getExpirationDate()->format('Y-m-d'));
+        $this->assertEquals('DVLA', $document->getIssuingAuthority());
+    }
+
```
bucky-boy: I know it's not in the Java SDK yet, but I wonder if we should add another test...
_shouldIgnoreAThridOptionalAttribute_
...and pass it a String like...
_'DRIVING_LICENCE GBR 1234abc 2016-05-01 DVLA someThirdAttribute'_

src/Yoti/Entity/DocumentDetails.php
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * @var string
+     */
+    private $issuingAuthority;
+
+    /**
+     * @var array
+     */
+    private $parsedValues = [];
```
bucky-boy: Do we need this here, or can it just be local to the _parseFromValue()_ method?

src/Yoti/Util/Profile/AttributeConverter.php
```diff
@@ -0,0 +1,27 @@
+<?php
+namespace Yoti\Util\Profile;
+
+
+use Yoti\Entity\Attribute;
+use Yoti\Entity\DocumentDetails;
+
+class AttributeConverter
+{
+    public static function convertValueBasedOnAttributeName($value, $attrName)
```
bucky-boy: Why not just pass in the whole attribute?

src/Yoti/ActivityDetails.php
```diff
@@ -4,10 +4,13 @@
 use Yoti\Entity\Selfie;
 use Yoti\Entity\Profile;
 use Yoti\Entity\Attribute;
+use Attrpubapi_v1\Attribute as ProtobufAttribute;
 use Attrpubapi_v1\AttributeList;
 use Yoti\Helper\ActivityDetailsHelper;
 use Yoti\Util\Age\AgeUnderOverProcessor;
 use Yoti\Util\Profile\AnchorProcessor;
+use Yoti\Util\Profile\AttributeConverter;
+use Yoti\Exception\AttributeException;
```
bucky-boy: Appears to be unused

src/Yoti/YotiClient.php
```diff
@@ -133,8 +134,8 @@ public function getOutcome()
      *
      * @return \Yoti\ActivityDetails
      *
-     * @throws \Exception
-     * @throws \Yoti\Exception\ActivityDetailsException
+     * @throws AttributeException
```
bucky-boy: I don't think we want it to throw AttributeException.
If there is a problem parsing a _DocumentDetails_ attribute, we'd want it to log a _warning_ (or the PHP equivalent) and continue as if the attribute was never there.

src/Yoti/ActivityDetails.php
sidibos: That's fine - I can change that.

src/Yoti/ActivityDetails.php
sidibos: I thought that could be confusing to users.
In this case the Attribute `document_details' is set in Dashboard and returned by connect, but because the value is invalid we are returning NULL. 
If we don't add it to the `$profileAttributes` map, the  user will think this attribute is not being sent by connect.
I am happy to not include it , if we think that's the way to go.

tests/Entity/DocumentDetailsTest.php
sidibos: I think this is going against the format we are expecting from the Document Details value.
As we are expecting 5 words not 6.

src/Yoti/Entity/DocumentDetails.php
sidibos: Yes it can, but I will have to pass the parsedValues as a param to each setter in that case.

src/Yoti/Util/Profile/AttributeConverter.php
sidibos: We don't need all the properties in the attribute Object, we only need those 2 params.

src/Yoti/ActivityDetails.php
sidibos: Yes I will remove that, I din't want to restrict the catch on `getYotiAttribute` to this exception only and forgot to remove it.

src/Yoti/YotiClient.php
sidibos: Yes the exception is fine, I will log the waring in `ActivityDetails::getYotiAttribute` inside the catch.

tests/Entity/DocumentDetailsTest.php
bucky-boy: True, but it doesn't hurt to prove that we won't fail if a 6th attribute starts appearing on DocumentDetails.  We would just ignore the 6th attribute, and process the first 3-5 as normal.

src/Yoti/Entity/DocumentDetails.php
bucky-boy: Probably cleaner to do that.  Makes the behaviour more obvious.

src/Yoti/YotiClient.php
```diff
@@ -131,10 +132,9 @@ public function getOutcome()
      *
      * @param string $encryptedConnectToken
      *
-     * @return \Yoti\ActivityDetails
+     * @return ActivityDetails
      *
-     * @throws \Exception
-     * @throws \Yoti\Exception\ActivityDetailsException
+     * @throws ActivityDetailsException
```
bucky-boy: So, should this be here?  Or should it be removed?

sidibos: If you are referring to `ActivityDetailsException` yes it should stay there.

tests/Entity/DocumentDetailsTest.php
sidibos: Ok - makes sense.

PR #37, opened 2018-10-24
SDK-577: Review AgeVerifications Implementations

sidibos:


src/Yoti/ActivityDetails.php
```diff
@@ -50,11 +46,6 @@ class ActivityDetails
      */
     private $pem;
 
-    /**
-     * @var \Yoti\Util\Age\Condition
-     */
-    private $ageCondition;
-
```
echarrod: Is this going to be a major release then, assuming this is a breaking change?

src/Yoti/ActivityDetails.php
```diff
@@ -206,225 +129,12 @@ public function getProfile()
...
-    }
-
-    /**
-     * Get user id.
+     * Get rememberMeId.
      *
      * @return string
      */
-    public function getUserId()
+    public function getRememberMeId()
```
echarrod: ditto - major release, right?

tests/Entity/ProfileTest.php
```diff
@@ -50,4 +67,91 @@ public function testGetAttributeName()
...
         $this->assertEquals('phone_number', $phoneNumber->getName());
     }
+
+    public function testGetAgeVerifications()
+    {
+        $profileData = [
+            'age_over:18' => new Attribute('age_over:18', 'true', [], []),
+            'age_under:18' => new Attribute('age_under:18', 'false', [], []),
+        ];
+        $processor = new Processor($profileData);
```
bucky-boy: Looks like we are testing _Processor_ rather than _Profile_.  This should be in its own test class.

tests/Entity/ProfileTest.php
```diff
@@ -50,4 +67,91 @@ public function testGetAttributeName()
...
+
+    public function testShouldReturnFormattedAddressAsPostalAddressWhenNull()
+    {
+        $structuredPostalAddress = new Attribute(
+            Profile::ATTR_STRUCTURED_POSTAL_ADDRESS,
+            $this->dummyStructuredPostalAddress,
+            [],
+            []
+        );
+        $prolieData = [
```
bucky-boy: profileData

tests/Entity/ProfileTest.php
```diff
@@ -50,4 +67,91 @@ public function testGetAttributeName()
...
+            json_encode($profile->getStructuredPostalAddress()->getValue())
+        );
+    }
+
+    /**
+     * Should not return age_verifications in the array
+     */
+    public function testGetAttributes()
+    {
+        $prolieData = [
```
bucky-boy: profileData?

src/Yoti/Entity/Profile.php
```diff
@@ -106,24 +112,53 @@ public function getStructuredPostalAddress()
...
+     *
+     * @return array
+     * e.g [
+     *      'age_under:18' => new AgeVerification(...),
+     *      'age_over:50' => new AgeVerification(...),
+     *      ...
+     * ]
      */
-    public function getAgeCondition()
+    public function getAgeVerifications()
```
bucky-boy: I understand that this method will return a collection of AgeVerification objects.
1. If there are no AgeVerification objects, maybe it would be better to return an empty list rather than _null_
2. We will still need to implement get-attribute-by-name in this Profile (and in the ApplicationProfile) i.e. a new method something like:
```
getAttribute(name) { ... }
```
When we do that, sdk users must still be able to call _getAttribute('age_over:18') and get the original attribute back (not the AgeVerification object).  Is that going to work with the approach in this PR?

src/Yoti/Util/Age/BaseAgeProcessor.php
```diff
@@ -0,0 +1,77 @@
...
+     *
+     * @param Attribute $attribute
+     *
+     * @return array
+     */
+    protected function extractAgeVerificationData(Attribute $attribute)
+    {
+        $ageCheckArr = explode(static::AGE_DELIMITER, $attribute->getName());
+        if (count($ageCheckArr) > 1) {
+            return [
```
bucky-boy: Why not just get this object to return an AgeVerification object?
Looks like everything we need is here already.
At present we return a map here.  Then in _Processor_ we use that map and the original Attribute to create an AgeVerification.  Why not just create the AgeVerification straight away, and save the hassle?

src/Yoti/Entity/Profile.php
```diff
@@ -106,24 +112,53 @@ public function getStructuredPostalAddress()
...
     {
-        return $this->getProfileAttribute(self::ATTR_AGE_CONDITION);
+        return isset($this->profileData[self::ATTR_AGE_VERIFICATIONS])
```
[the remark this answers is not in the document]
sidibos: Well `isset` is used on the others as well, but there is an extra check conducted on them, please see BaseProfile - line 26

src/Yoti/Util/Age/BaseAgeProcessor.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace Yoti\Util\Age;
+
+use Yoti\Entity\Attribute;
+
+class BaseAgeProcessor implements AgeProcessorInterface
+{
+    // You must define this value in the child class
+    const AGE_DELIMITER = '';
```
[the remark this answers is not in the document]
sidibos: Not sure about that, the delimiter might change, i prefer to leave the options opened.
This is overridden in the child class.

src/Yoti/Entity/Profile.php
sidibos: 1. We want to be consistent with the other getters, if the attribute isn't set in the Dashboard we return NULL.

2. getAttribute(name){} is already implemented in BaseProfile::class and both Profiles class are extending this.

tests/Entity/ProfileTest.php
sidibos: Yep good catch!

src/Yoti/Util/Age/BaseAgeProcessor.php
```diff
@@ -0,0 +1,77 @@
...
+
+    // You must define this pattern in the child class
+    const AGE_RULE_PATTERN = '';
```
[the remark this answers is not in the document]
sidibos: If it's declared Abstract, it won't have any abstract function, So no really need to be abstract.

src/Yoti/Util/Age/AgeProcessorInterface.php
```diff
@@ -0,0 +1,7 @@
+<?php
+namespace Yoti\Util\Age;
+
+interface AgeProcessorInterface
```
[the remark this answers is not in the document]
sidibos: yes but don't forget that this method can be overridden by any child class, depending on the complexity of the process.
Also this helps the processor to dynamically invoke the process function.
Please see[ here](https://github.com/getyoti/yoti-php-sdk/blob/SDK-577/src/Yoti/Util/Age/Processor.php#L53).

src/Yoti/Util/Age/Processor.php
```diff
@@ -2,53 +2,68 @@
 
 namespace Yoti\Util\Age;
 
+use Yoti\Entity\AgeVerification;
+
 class Processor
```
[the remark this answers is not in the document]
sidibos: Are you fine with AgeProcessor?

tests/Entity/ProfileTest.php
sidibos: It does look like, because we don't have a sample receipt with age derivations.
I am happy to move it inside ProcessorTest.

src/Yoti/Util/Age/BaseAgeProcessor.php
sidibos: As you wish!

src/Yoti/Entity/Profile.php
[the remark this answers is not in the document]
sidibos: Also I am happy to add these two helpers to the Profile:
Profile::findAgeOverVerification(int age)
Profile::findAgeUnderVerification(int age)

src/Yoti/Util/Age/BaseAgeProcessor.php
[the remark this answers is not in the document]
bucky-boy: If it changes, that would be a breaking change, and everything will fall over.  It's specified as a ':' in the docs.  Honestly, it's not changing.

src/Yoti/Util/Age/BaseAgeProcessor.php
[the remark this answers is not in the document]
bucky-boy: BaseAgeProcessor itself has no meaning.  It does not actually know enough to process any kind of age-derived attribute.  Instantiating a BaseAgeProcessor makes no sense - you're expected to subclass it and fill in the missing bits.  So it is in every sense an abstract class.

If you really want one, your abstract function could be _getAgePattern()_ - that's the bit of functionality that subclasses are really expected to provide.

src/Yoti/Util/Age/AbstractAgeProcessor.php
```diff
@@ -2,40 +2,85 @@
...
-     * Get age attribute and value.
+     * Return true if attribute name matches the pattern.
      *
-     * @return null|array
+     * @param Attribute $attribute
+     *
+     * @return bool
      */
-    public function getAgeRow()
+    protected function isDerivedAttribute(Attribute $attribute)
```
bucky-boy: Would be clearer if it was called _handlesAttribute_ or _processesAttribute_

src/Yoti/Util/Age/AbstractAgeProcessor.php
```diff
@@ -2,40 +2,85 @@
 
 namespace Yoti\Util\Age;
 
+use Yoti\Entity\Attribute;
+use Yoti\Entity\AgeVerification;
+
 abstract class AbstractAgeProcessor
```
bucky-boy: We now have both an _AbstractAgeProcessor_ and an _AgeProcessor_.
Might be a little clearer if this one were to become _AgeVerificationFactory_ (or _BaseAgeVerificationFactory_, or _AgeVerificationConverter_ etc) and the two sublasses would be _AgeOverVerificationFactory_ + _AgeUnderVerificationFactory_.

src/Yoti/Util/Age/AbstractAgeProcessor.php
sidibos: I would prefer to keep it as it is, as this is returning a boolean so that name reflects what it's doing.

src/Yoti/Util/Age/AbstractAgeProcessor.php
sidibos: I see your point, I will call it _AgeVerificationConverter_, but I will rather call the two subclasses _AgeOverVerificationProcessor_ and _AgeUnderVerificationProcessor_ as firstly it's not only creating it's also processing data, secondly the name is inline with the _AbstractAgeProcessor_.

PR #41, opened 2018-11-07
SDK-666: Include SDK version in the request headers

sidibos:
- Include SDK version in the Request Header
- Add Config class to retrieve config variables
- Add Test case for Config::class and SDK version inclusion in the request headers.


tests/Util/ConfigTest.php
```diff
@@ -0,0 +1,20 @@
...
+class ConfigTest extends TestCase
+{
+    public function testConfigInstance()
+    {
+        $this->assertInstanceOf(Config::class, Config::getInstance());
+    }
+
+    public function testGetSDKVersion()
+    {
+        $sdkVerision = Config::getInstance()->get('version');
```
qzagarese: typo

tests/YotiClientTest.php
```diff
@@ -172,6 +172,24 @@ public function testCanUsePHPAsSdkIdentifier()
...
 
+    public function testRequestHeadersShouldIncludeSDKVersion()
+    {
+        $yotiClientObj  = new YotiClient(
+            SDK_ID,
+            $this->pem,
+            YotiClient::DEFAULT_CONNECT_API,
+            'PHP'
+        );
+        $requestHeaders = $this->invokeMethod(
```
qzagarese: Same consideration as for https://github.com/getyoti/yoti-php-sdk/pull/40 applies.
If possible, try to avoid usage of reflection to test methods.
You can try to mock whatever is performing the http request and verify the values of the headers in the request.

sidibos: This has been addressed in [SDK-658](https://github.com/getyoti/yoti-php-sdk/blob/SDK-658/tests/Http/CurlRequestHandlerTest.php#L80), so will be reflected here once merged.

tests/Util/ConfigTest.php
sidibos: Will fix that - thx.

PR #42, opened 2018-11-15
SDK-659: Add the ability to retrieve a list of Attribute Anchors

sidibos:


src/Yoti/Entity/Anchor.php
```diff
@@ -4,33 +4,43 @@
 
 class Anchor
 {
+    const TYPE_SOURCE_OID = '1.3.6.1.4.1.47127.1.1.1';
```
bucky-boy: We shouldn't need these two constants to be defined in two places

src/Yoti/Util/Profile/AnchorConverter.php
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+    }
+
+    /**
+     * @param string $extEncodedValue
+     *
+     * @return null
+     */
+    private function decodedAnchorValue($extEncodedValue)
```
bucky-boy: Small thing - _decodeAnchorValue_ was correct.  Method names are better as verbs

src/Yoti/Util/Profile/AnchorConverter.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * Convert Protobuf Anchors to a map of Yoti Anchors
+     * And merge the result into the reference array
+     *
+     * @param Anchor $anchor
+     * @param $yotiAnchorsMap
+     */
+    public function convertToYotiAnchors(Anchor $anchor, &$yotiAnchorsMap)
```
bucky-boy: This could be a little simpler...

_convertToYotiAnchors_ - this method works on one Anchor at a time, so the method name should be singular

It would be a bit cleaner if this method simply returned the new _Anchor_, rather than adding it to the map.  Let the _AnchorListConverter_ manage the collection.
So then we would have _AnchorConverter_, which just converts _Anchors_ one at a time, and separately we'd have _AnchorListConverter_ which converts a list of Anchors into the Map.

src/Yoti/Util/Profile/AnchorConverter.php
```diff
@@ -0,0 +1,117 @@
...
+     * @param \Attrpubapi_v1\Anchor $anchor
+     *
+     * @return \Yoti\Entity\SignedTimeStamp
+     */
+    private function convertToYotiSignedTimestamp(Anchor $anchor)
+    {
+        $signedTimeStamp = new \Compubapi_v1\SignedTimestamp();
+        $signedTimeStamp->mergeFromString($anchor->getSignedTimeStamp());
+
+        $timeInSeconds = round($signedTimeStamp->getTimestamp()/1000000);
```
bucky-boy: The timestamp we return has to be accurate to the microsecond.
Also, what TimeZone will be set on this?  Everything has to be UTC.

Java _time_ classes don't do microsecond accuracy so we had to create a new object for it.  That was a pain, but does have the advantage that we don't have to consider time zones anymore - we simply never included a timezone on it and documented that everything is UTC.  We don't use any of _time_ classes, so we don't have to worry about time zone conversion.

src/Yoti/Util/Profile/AnchorListConverter.php
```diff
@@ -0,0 +1,22 @@
...
+namespace Yoti\Util\Profile;
+
+use Traversable;
+
+class AnchorListConverter
+{
+    public static function convert(Traversable $anchorList)
+    {
+        $yotiAnchorsMap = [];
+        $anchorConverter = new AnchorConverter();
```
bucky-boy: If _AnchorConverter_ is threadsafe then we don't need to create a new one every time we enter this method.

src/Yoti/Util/Profile/AnchorConverter.php
sidibos: Good spot - That slips out of my mind.

src/Yoti/Util/Profile/AnchorConverter.php
sidibos: The thing is each protobuf anchor may result to more than one YotiAnchor. As YotiAnchors are produced for each protobuf Anchor extensions.

src/Yoti/Util/Profile/AnchorListConverter.php
sidibos: Unfortunately no - The ASN1 and X509 object could be overwritten, so I will leave it as it is.

src/Yoti/Entity/Anchor.php
sidibos: Which two places?

bucky-boy: Anchor and AnchorProcessor

src/Yoti/Util/Profile/AnchorConverter.php
bucky-boy: Do you mean that because there's a list of certificates in the AnchorProto there could be multiple Anchors coming out?

It's not necessarily clear from the structure, but we can be confident that one AnchorProto will result in one Anchor.  There are some notes in SDK-397, the ticket for the original implementation.

src/Yoti/Entity/Anchor.php
sidibos: There is no ore AnchorProcessor - it's been deleted !

src/Yoti/Util/Profile/AnchorConverter.php
sidibos: Timestamp seems to be in UTC `12-07-2017 20:24:40 UTC`
Will double check anyway.

src/Yoti/Util/Profile/AnchorConverter.php
sidibos: You may have two verifiers on a single attribute. e.g my address has two verifiers

Verifier YOTI_IDENTITY
Verifier YOTI_ADMIN

src/Yoti/Util/Profile/AnchorConverter.php
```diff
@@ -0,0 +1,172 @@
...
+use Traversable;
+use phpseclib\File\ASN1;
+use phpseclib\File\X509;
+use Attrpubapi_v1\Anchor;
+use Yoti\Entity\Anchor as YotiAnchor;
+
+class AnchorConverter
+{
+    /**
+     * Convert Protobuf Anchor to a map of Yoti Anchor
```
echarrod: Think this could be worded a bit differently, maybe:
`Convert Protobuf Anchor to a map of oid -> Yoti Anchor` ?

src/Yoti/Util/Profile/AnchorListConverter.php
```diff
@@ -0,0 +1,20 @@
...
+
+class AnchorListConverter
+{
+    public static function convert(Traversable $anchorList)
+    {
+        $yotiAnchorsMap = [];
+
+        foreach ($anchorList as $protobufAnchor) {
+            if ($parsedAnchor = AnchorConverter::convert($protobufAnchor)) {
+                $yotiAnchorsMap[$parsedAnchor['oid']][] = $parsedAnchor['yoti_anchor'];
```
echarrod: Will this allow us to have multiple sources and multiple verifiers with the same oid (key) for an attribute?

sidibos: Yes and there is a test case for that [here](https://github.com/getyoti/yoti-php-sdk/blob/SDK-659/tests/Util/Profile/AnchorListConverterTest.php#L25)

src/Yoti/Util/Profile/AnchorConverter.php
sidibos: I am not sure a bout using a long name for a function.

echarrod: I was referring to the doc comments 👍

src/Yoti/Util/Profile/AnchorListConverter.php
echarrod: 👍

src/Yoti/Util/Profile/AnchorConverter.php
sidibos: Ok sure.

src/Yoti/Entity/Attribute.php
```diff
@@ -70,4 +77,54 @@ public function getVerifiers()
...
+    {
+        $this->verifiers = $this->getAnchorType(
+            $anchorsMap,
+            Anchor::TYPE_VERIFIER_OID
+        );
+    }
+
+    private function setAnchors(array $anchorsMap)
+    {
+        // Remove Oids from the anchorsMap
```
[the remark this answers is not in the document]
sidibos: The alternative will be to loop through all the arrays filtering by type for each type, which is slower than the current implementation.

sidibos:
We could point example to the `src` of the current folder, but this will affect the sync in the number of SDK downloads and may hide the fact they could use composer to include Yoti-SDK as a dependency.
Also examples folder doesn't have the coder SDK component, always look into `src` and `tests` folders.
If you want it can added in the wiki.

PR #44, opened 2018-12-12
SDK-603: Add Sandbox Support

sidibos:


sandbox/src/Http/RequestBuilder.php
```diff
@@ -0,0 +1,217 @@
...
+    {
+        $this->rememberMeId = $value;
+    }
+
+    public function setFullName($value, $optional = 'false', array $anchors = [])
+    {
+        $this->addAttribute($this->createAttribute(
+            Profile::ATTR_FULL_NAME,
+            $value,
+            '',
```
echarrod: Since PHP doesn't have named parameters, do you think it would be useful to set `$noDerivation = ''`, and then use `$noDerivation` (or similar) in each of these calls to createAttribute? I think it would be helpful when looking at the code, rather than having to navigate to `createAttribute` to see what this argument is

sandbox/src/SandboxClient.php
```diff
@@ -0,0 +1,107 @@
...
+     * @param RequestBuilder $requestBuilder
+     *
+     * @param string $endpoint
+     * @param string $httpMethod
+     *
+     * @return array
+     *
+     * @throws \Yoti\Exception\RequestException
+     */
+    protected function sendRequest(RequestBuilder $requestBuilder, $endpoint, $httpMethod)
```
bucky-boy: I don't think this will ever be overridden.  It could be private.

sandbox/src/Http/RequestBuilder.php
```diff
@@ -0,0 +1,217 @@
...
+        $this->addAttribute($this->createAttribute(
+            Profile::ATTR_STRUCTURED_POSTAL_ADDRESS,
+            $value,
+            '',
+            $optional,
+            $anchors
+        ));
+    }
+
+    public function setDocumentDetails($value, $optional = 'true', array $anchors = [])
```
bucky-boy: I'm not clear what type the _$value_ parameter should be.

We need to provide users with a fool-proof way to create this test value - I don't want to trust them to give us a correctly formatted string.  Ideally, we give them an easy, safe way to build a well-formed _DocumentDetails_ instance and then they can pass that to us.  We will convert that object to a string for them, because
1. we know what that format should be
2. if there is any problem with the structure, we can check it before we even bother sending it to _connect_.  Should be easier to identify problems that way.

PR #48, opened 2019-03-04
SDK-810: E2E Tests

davidgrayston:
- Allowing API URLs to be configured


examples/.env.dist
```diff
@@ -5,4 +5,9 @@ YOTI_APP_ID=xxxxxxxxxxxxxxxxxx
 YOTI_SCENARIO_ID=xxxxxxxxxxxxxxxx
 YOTI_SDK_ID=xxxxxxxxxxxxxxxxxxxxx
 # Below is the private key (in .pem format) associated with the Yoti Application you created on Dashboard
-YOTI_KEY_FILE_PATH=/path_to_pem_dir/php-sdk-access-security.pem
\ No newline at end of file
+YOTI_KEY_FILE_PATH=/path_to_pem_dir/php-sdk-access-security.pem
+
+# Change the API URLs (optional)
```
echarrod: I'm just thinking whether we could put these new env values in a separate .env file. Clients might be a little overwhelmed when they see all these config values, when in all likelihood they won't be using these optional ones ever. @KiranBali what do you think?

davidgrayston: We could put in `.env.dev.dist` if that's appropriate?

echarrod: Yes that would be good 👍

davidgrayston: Moved to `.env.dev`

PR #52, opened 2019-03-13
SDK-766: Document Images

davidgrayston:
> Note: It might be clearer looking at changes separately from regeneration of protobuf files:
[e5224925bb05d69fe46c2d6fe696b37eb0413e89...SDK-766-document-images](https://github.com/getyoti/yoti-php-sdk/compare/e5224925bb05d69fe46c2d6fe696b37eb0413e89...SDK-766-document-images)

### MultiValue object
All `MULTI_VALUE` attribute values are converted to `MultiValue` objects:
- Allows recursive lists of any type (tree)
- Tree can be filtered by instance of or type
- Extends [ArrayObject](http://php.net/manual/en/class.arrayobject.php) so behaves like an array (implements `IteratorAggregate` , `ArrayAccess` , `Serializable` , `Countable`)
- Empty values are added to the list as `NULL` (and logged as per existing empty attribute values)

#### Methods:
- `::filterInstance()` allows filtering by the instance
- `::filterType()` allows filtering by value returned from [gettype()](http://php.net/manual/en/function.gettype.php)
- `::filter()` allows custom filtering (used by `::filterInstance()` and `::filterType()`)
- `::resetFilters()` resets filters and values

#### Example usage:
```php
$multiValue = new MultiValue($items);
$multiValue
  ->filterInstance(ImageJpeg::class)
  ->filterType('string');
```

### Document Images
- Added `Profile::getDocumentImages()`

  Returns an array of `Image` objects using:
  ```php
  $multiValue
    ->filterInstance(Image::class)
    ->getArrayCopy();
  ```

### Protobuf
- Regenerated files with new MultiValue content type


src/Yoti/Util/Profile/AttributeConverter.php
```diff
@@ -8,13 +8,15 @@
...
 use Yoti\Exception\AttributeException;
+use Yoti\Entity\MultiValue;
 
 class AttributeConverter
 {
     const CONTENT_TYPE_JSON = 5;
     const CONTENT_TYPE_PNG = 4;
     const CONTENT_TYPE_JPEG = 2;
     const CONTENT_TYPE_DATE = 3;
+    const CONTENT_TYPE_MULTI_VALUE = 6;
```
echarrod: I thought these could be alphabetical, but doesn't seem like there's any order. Could we change them to order by the enum value?

src/Yoti/Util/Profile/AttributeConverter.php
```diff
@@ -65,18 +69,50 @@ private static function convertValueBasedOnContentType(ProtobufAttribute $protob
...
             case self::CONTENT_TYPE_DATE:
                 $value = self::convertTimestampToDate($value);
                 break;
+
+            case self::CONTENT_TYPE_MULTI_VALUE:
+                $value = self::convertMultiValue($value);
+                break;
         }
 
         return $value;
```
echarrod: Could we also add `case self::CONTENT_TYPE_STRING:`, and also for `CONTENT_TYPE_UNDEFINED` here, (and add them to the constants at the top). That way it is clearer that they're not just cases that we haven't considered, and it's clear what we are doing with them.

tests/Util/Profile/AttributeConverterTest.php
```diff
@@ -7,16 +7,25 @@
...
  */
 class AttributeConverterTest extends TestCase
 {
+    /**
+     * Content Types.
+     */
+    const CONTENT_TYPE_STRING = 0;
+    const CONTENT_TYPE_PNG = 4;
+    const CONTENT_TYPE_JPEG = 2;
+    const CONTENT_TYPE_MULTI_VALUE = 6;
```
echarrod: Same here RE ordering of vars

src/Yoti/Util/Profile/AttributeConverter.php
davidgrayston: I've reorganised the switch statement to include `CONTENT_TYPE_STRING`/`CONTENT_TYPE_UNDEFINED` and return values.

tests/Util/Profile/AttributeConverterTest.php
davidgrayston: Fixed in 152f09b

src/Yoti/Util/Profile/AttributeConverter.php
davidgrayston: Fixed in 152f09b

davidgrayston:
@echarrod I've made Document Images immutable in 199b530 - `getDocumentImages()` now returns the immutable `MultiValue` object, that behaves like an array.

I've also added test extra coverage to check sample image content is correct as per Go and Node SDKs.

davidgrayston:
@echarrod in 04b28a9 I'm allowing empty strings in MultiValues.

This change also allows empty string attribute values, but does not allow empty non-string values.

>Note: This covers SDK-859

src/Yoti/Entity/MultiValue.php
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        // Reset to original items before filtering.
+        $this->exchangeArray($this->originalItems);
+
+        // Apply filter to items.
+        if (count($this->filters) > 0) {
+            $filtered_array = array_filter(
+                $this->getArrayCopy(),
+                [$this, 'filterItem']
```
qzagarese: Is this reflective call necessary?
Can't we replace this with something that is checked at load time?

davidgrayston: I can change this to an [anonymous function / closure](http://php.net/manual/en/functions.anonymous.php)? It might make more sense in this case

davidgrayston: I've changed this in 7f0912e0e594ad6f031b4bef55ea0ba4c7d7f1bf

PR #58, opened 2019-04-02
SDK-896: Check log messages are correct

davidgrayston:
Following #57 I've added test coverage for the logged warnings.

>Note: I'll close this PR if we feel this is unnecessary.

### DIFF
--- a/src/Yoti/Util/Profile/AttributeConverter.php
+++ b/src/Yoti/Util/Profile/AttributeConverter.php
@@ -9,6 +9,7 @@ use Compubapi\EncryptedData;
 use Attrpubapi\Attribute as ProtobufAttribute;
 use Yoti\Exception\AttributeException;
 use Yoti\Entity\MultiValue;
+use Yoti\Entity\ApplicationProfile;
 
 class AttributeConverter
 {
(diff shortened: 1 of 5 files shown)


davidgrayston:
@echarrod I've addressed logging issues in 8ac93a9
- Unexpected logs will now appear in CI output
- No longer logging when application logo is empty as this is valid (not an exception) - I've added the check before regular attribute conversion, as the logic depends on both content type **and** attribute name
- Added extra test coverage for expected log messages

src/Yoti/Util/Profile/AttributeConverter.php
```diff
@@ -166,6 +166,12 @@ public static function convertToYotiAttribute(ProtobufAttribute $protobufAttribu
     {
         $yotiAttribute = null;
 
+        // Application Logo can be empty, return NULL when this occurs.
+        if ($protobufAttribute->getName() == Profile::ATTR_APPLICATION_LOGO &&
```
echarrod: We have separate constants for UserProfile and ApplicationProfile, you can find the application_logo const [here](https://github.com/getyoti/yoti-php-sdk/blob/master/src/Yoti/Entity/ApplicationProfile.php#L7).

davidgrayston: Good spot 👍 fixed in 3655638

Issue #61, opened 2019-05-01
Insufficient checks for PEM file existence 


HARVS1789UK:
Seen in a recent (2019-04-30) composer installation of getyoti/yoti-php-sdk at version `2.2.0` 

`Yoti\YotiClient::extraxctPemContent()` takes a file path, checks the file exists and validates that the file contains a valid private key.

The following lines are responsible for checking the file exists and if so updating the `$pem` variable* to contain the private key data.

```
        // Check that file exists if user passed PEM as a local file path
        if (strpos($pem, 'file://') !== false && !file_exists($pem)) {
            throw new YotiClientException('PEM file was not found.', 400);
        }

        // If file exists grab the content
        if (file_exists($pem)) {
            $pem = file_get_contents($pem);
        } 
``` 

I have a few questions and issues here:

- 1 - With regards to the comment with reference to 'local file path', I appreciate that (to quote from it's docs) file_exists() can:

>As of PHP 5.0.0, this function can also be used with some URL wrappers.

but in the context of a downloaded .pem file to be stored in your applications code base, is anything **but** a local path likely to be used?

- 2 - I don't know if this is a Windows thing (I don't think so?) but there is no need to prefix a local file path with 'file://' to my knowledge and the presence of that additional check means any local file path in the form of, for example `/troublesome/absolute/path/to/my/file.pem` will fail to throw the "PEM file was not found" exception and instead will throw "PEM key content is invalid" once the file path is passed to `openssl_get_privatekey()` (as that will still be what `$pem` contains, another good reason not to use the same, passed by reference variable to hold the file path and file contents).
This is pretty misleading and leads to unnecessary additional debugging when you accidentally specify the wrong local file path

- 3 - What is the purpose of the second call to `file_exists()` on line 350?
If the issues in No. 2 above were remedied it would be unnecessary but if it is required, surely it too should have an `else {}` condition which throws the same "PEM file was not found" exception?

Certainly I feel that the `file_exists()` logic here needs some changes to prevent misleading error messaging, and potentially the entire method could be modified to be a little clearer, perhaps as per my suggestion below:

```
    /**
     * Validate and store PEM file content.
     *
     * @param string $pem - An absolute path to the PEM file containing the Yoti
     * applications private key
     * @throws YotiClientException
     * @return void
     */
    private function extractPemContent($pemPath)
    {
        // Check PEM path is provided
        if (!$pemPath) {
            throw new YotiClientException('PEM file path is required', 400);
        }

[...]

PR #62, opened 2019-05-02
SDK-934: PEM file handling

davidgrayston:
Relates to #61 

- Improving PHPDocs to cover PEM content being passed as both file path and string
- Added further test coverage for PEM strings and files


ZihaoT:
LGTM

PR #65, opened 2019-07-08
SDK-1039: Modal QR

davidgrayston:
### Changed
- Example project now uses modal QR
- Example `composer.json` will now fetch any `2.*` version of the SDK 
- Updated Example lock file with latest stable tag
- Running the Example with `docker-compose` will use the local checked out SDK
- Using official composer Docker container

### Added
- Example now supports structured postal address


echarrod:
I've added the tickets under the SDK-493 epic

PR #69, opened 2019-07-17
SDK-1039: Allow composer install without Docker

davidgrayston:
In #65 I updated the examples to use the local checked out SDK, however, this isn't working correctly outside docker (running `composer install` on host machine)

Composer doesn't allow parent directories as a repository, e.g.
```
  "repositories": [
    {
      "type": "path",
      "url": "../",
      "options": {
        "symlink": true
      }
    }
  ],
```

As a result I've changed the implementation to:
- Copy the parent directory into `./examples/sdk` _(excluding `./examples`)_
- `composer.json` will then install the SDK from `./examples/sdk`

>I'd appreciate any suggestions on a more elegant way of implementing `"copy-sdk"` script in `composer.json`

I've also updated the SDK version and the `./examples/composer.lock` to reflect this. 

This should work with both Docker and on host machine:
- `composer update` / `composer install`
- `docker-compose up -d` or specifically: `docker-compose up composer`

### DIFF
--- a/examples/composer.lock
+++ b/examples/composer.lock
@@ -67,10 +67,9 @@
             "name": "yoti/yoti-php-sdk",
             "version": "2.2.0",
             "dist": {
-                "type": "zip",
-                "url": "https://api.github.com/repos/getyoti/yoti-php-sdk/zipball/f3c770c1df2cbba3dc76a1f329938c379191ad7a",
-                "reference": "f3c770c1df2cbba3dc76a1f329938c379191ad7a",
-                "shasum": ""
+                "type": "path",
+                "url": "./sdk",
+                "reference": "549162d0d6f3c0e2b1c328f9128fb28c715da103"
             },
             "require": {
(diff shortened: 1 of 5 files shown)


examples/composer.lock
```diff
@@ -95,11 +94,21 @@
...
+                "psr-4": {
+                    "YotiTest\\": "tests/",
+                    "SandboxTest\\": "sandbox/tests/"
+                }
+            },
             "license": [
                 "MIT"
             ],
             "authors": [
+                {
```
echarrod: Do you know why these lines were moved? Would be clearer if they stayed in the same place!

davidgrayston: I'm not sure why this moved, I'll revert to see if this is consistent behaviour- It's now the same order as `composer.json` so seems more correct this way

davidgrayston: As discussed, I've removed authors in 1c1020e

PR #75, opened 2019-08-23
SDK-1080: Signed Request Builder

davidgrayston:
### Example Usage
```php
use Yoti\Http\RequestBuilder;
use Yoti\Http\Payload;

$request_handler = (new RequestBuilder)
  ->withBaseUrl($base_url)
  ->withPemString($pem_string)
  ->build();

$get_result = $request_handler->get(
  '/some-endpoint',
  ['some' => 'value']
);

$post_result = $request_handler->post(
  '/some-endpoint',
  new Payload(['some' => 'payload']),
  ['some' => 'value']
);
```

> Note: `Yoti\Http\Payload` is a pre-existing class that also supports payload strings

#### Optional builder methods
```php
$request_handler = (new RequestBuilder)
  ->withBaseUrl($apiUrl)
  ->withPemFilePath('/path/to/file.pem') // File path
  ->withSdkIdentifier($sdkIdentifier) // To allow plugins to override identifier header
  ->withSdkVersion($sdkVersion) // To allow plugins to override version header
  ->withHeader('Custom-Header', 'some-value') // To allow additional headers
  ->build();
```

### Added
- `Yoti\Http\RequestBuilder`
- `Yoti\Util\PemFile`
- `Yoti\Exception\PemFileException`
- `Yoti\Http\AbstractRequestHandler::get()`
- `Yoti\Http\AbstractRequestHandler::post()`

### Deprecated
- `Yoti\Http\AbstractRequestHandler::getSdkId()` - this will no longer be required (`appId` query param is added by `YotiClient` when making requests)

### Changed
- SDK version can now be overridden - helps resolve SDK-1071

### Fixed
- Updated activity details timestamp test to use UTC timezone when testing date format (failed when default timezone was not `UTC`, such as `Europe/Zurich`)
- Updated example project Dockerfile for `nginx:latest` upgrade to Debian Buster


src/Yoti/Http/AbstractRequestHandler.php
```diff
@@ -43,50 +47,106 @@ abstract class AbstractRequestHandler
...
+     *
+     * @throws RequestException
+     */
+    public function post($endpoint, Payload $payload = null, array $queryParams = [])
+    {
+        return $this->sendRequest($endpoint, self::METHOD_POST, $payload, $queryParams);
+    }
+
+    /**
+     * @deprecated will be removed in version 3 - SDK ID is now added as a query param.
```
echarrod: Would be good if we said _where_ it's added as a query param too, since it's not in this file 👍

src/Yoti/Http/AbstractRequestHandler.php
```diff
@@ -98,61 +158,72 @@ public function getSdkId()
...
-            CurlRequestHandler::YOTI_DIGEST_HEADER_KEY . ": {$signedMessage}",
-            CurlRequestHandler::YOTI_SDK_IDENTIFIER_KEY . ": {$this->sdkIdentifier}",
-            'Content-Type: application/json',
-            'Accept: application/json',
-        ];
-
-        if ($version = Config::getInstance()->get('version')) {
-            $requestHeaders[] = self::YOTI_SDK_VERSION . ": {$this->sdkIdentifier}-{$version}";
+        if (!in_array($sdkIdentifier, $this->acceptedsdkIdentifiers, true)) {
+            throw new RequestException("Wrong Yoti SDK identifier provided: {$sdkIdentifier}");
```
echarrod: I'm not a fan of saying "Wrong" in error messages without saying why 🙂 
```suggestion
            throw new RequestException("{$sdkIdentifier} is not in the list of accepted identifiers");
```

tests/Http/RequestBuilderTest.php
```diff
@@ -0,0 +1,135 @@
...
+          ->build();
+
+        $this->assertInstanceOf(AbstractRequestHandler::class, $requestHandler);
+    }
+
+    /**
+     * @covers ::build
+     *
+     * @expectedException \Yoti\Exception\RequestException
+     * @expectedExceptionMessage Wrong Yoti SDK identifier provided: Invalid SDK
```
echarrod: If we updated the error then obvs would need to update here 👍

davidgrayston: Fixed in 2a8784b

src/Yoti/Http/AbstractRequestHandler.php
davidgrayston: Fixed in 2a8784b

src/Yoti/Http/AbstractRequestHandler.php
davidgrayston: Fixed in 2a8784b / 1aacef5

davidgrayston:
@echarrod as discussed, I've made `RequestBuilder::withPemFile()` private in 71aa430

davidgrayston:
@echarrod I've added support for custom headers in beaf8a8

PR #77, opened 2019-09-04
SDK-1080: Signed Request

davidgrayston:
> Revision of #75 - change log notes below include changes on #75

### Example Usage
```php
use Yoti\Http\RequestBuilder;
use Yoti\Http\Payload;

$request = (new RequestBuilder)
  ->withBaseUrl('https://somehost.com/api/v1')
  ->withEndpoint('/some-endpoint')
  ->withPemString($somePemString)
  ->withMethod('POST')
  ->withPayload(new Payload(['some' => 'payload']))
  ->build();

$request->getUrl(); // Returns string
$request->getMethod(); // Returns string
$request->getHeaders(); // Returns associative array of name => value
$request->getPayload(); // Returns \Yoti\Http\Payload
$request->setHandler($handler) // Expects \Yoti\Http\RequestHandlerInterface
$request->execute(); // Executes the request using default (Curl) or custom handler
```

> Note: `Yoti\Http\Payload` is a pre-existing class that also supports payload strings

#### Optional builder methods
```php
$request_handler = (new RequestBuilder())
  ...
  ->withQueryParam('some', 'param')
  ->withPemFilePath('/path/to/file.pem') // File path
  ->withSdkIdentifier('some-sdk-identifier') // To allow plugins to override identifier header
  ->withSdkVersion('some-sdk-version') // To allow plugins to override version header
  ->withHeader('Some-Header', 'some-value') // To allow additional headers
  ->withHandler($handler) // Expects \Yoti\Http\RequestHandlerInterface
  ->withPost() // Shortcut for ::withMethod()
  ->withGet() // Shortcut for ::withMethod()
  ->build();
```

### Added
- `Yoti\Http\RequestBuilder`
- `Yoti\Http\Request`
- `Yoti\Http\Response`
- `Yoti\Util\PemFile`
- `Yoti\Exception\PemFileException`
- `\Yoti\Http\RequestHandlerInterface`
- `\Yoti\Http\Curl\RequestHandler`
- `\Yoti\YotiClient::setRequestHandler()` - allows Curl handler to be replaced with an alternative such as [Guzzle](https://github.com/guzzle/guzzle)

### Deprecated
- `Yoti\Http\AbstractRequestHandler` - replaced by `\Yoti\Http\RequestHandlerInterface`
- `Yoti\Http\CurlRequestHandler` - replaced by `\Yoti\Http\Curl\RequestHandler`

### Changed
- SDK version can now be overridden - helps resolve SDK-1071

### Fixed
- Default Curl request handler now verifies host and peer certificates

### DIFF
--- a/src/Yoti/Http/AbstractRequestHandler.php
+++ b/src/Yoti/Http/AbstractRequestHandler.php
@@ -2,91 +2,115 @@
 
 namespace Yoti\Http;
 
-use Yoti\Util\Config;
 use Yoti\Util\PemFile;
 use Yoti\Exception\RequestException;
 
 abstract class AbstractRequestHandler
 {
-    // HTTP methods
-    const METHOD_GET = 'GET';
-    const METHOD_POST = 'POST';
-    const METHOD_PUT = 'PUT';
-    const METHOD_PATCH = 'PATCH';
-    const METHOD_DELETE = 'DELETE';
+    /**
+     * HTTP methods
+     *
+     * @deprecated 3.0.0 replaced by \Yoti\Http\Request
+     */
+    const METHOD_GET = Request::METHOD_GET;
+    const METHOD_POST = Request::METHOD_POST;
+    const METHOD_PUT = Request::METHOD_PUT;
+    const METHOD_PATCH = Request::METHOD_PATCH;
+    const METHOD_DELETE = Request::METHOD_DELETE;
 
-    // Request HttpHeader keys
-    const YOTI_AUTH_HEADER_KEY = 'X-Yoti-Auth-Key';
-    const YOTI_DIGEST_HEADER_KEY = 'X-Yoti-Auth-Digest';
-    const YOTI_SDK_IDENTIFIER_KEY = 'X-Yoti-SDK';
-    const YOTI_SDK_VERSION = 'X-Yoti-SDK-Version';
+    /**
+     * Request HttpHeader keys
+     *
+     * @deprecated 3.0.0 replaced by \Yoti\Http\RequestBuilder
+     */
+    const YOTI_AUTH_HEADER_KEY = RequestBuilder::YOTI_AUTH_HEADER_KEY;
+    const YOTI_DIGEST_HEADER_KEY = RequestBuilder::YOTI_DIGEST_HEADER_KEY;
+    const YOTI_SDK_IDENTIFIER_KEY = RequestBuilder::YOTI_SDK_IDENTIFIER_KEY;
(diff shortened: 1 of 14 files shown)


src/Yoti/Http/AbstractRequestHandler.php
```diff
@@ -2,158 +2,117 @@
...
-     *
-     * @param string $endpoint
-     * @param Payload|NULL $payload
-     * @param array queryParams
-     *
-     * @return array
-     *
-     * @throws RequestException
-     */
-    public function post($endpoint, Payload $payload = null, array $queryParams = [])
```
emmas-yoti: When was this introduced? Is this a breaking change or was this not released?

davidgrayston: This is in development but not released (was in initial signed request implementation)

davidgrayston: Same applies to `::get()` that's been removed in this PR

davidgrayston:
You can use Guzzle by implementing the RequestHandler interface like this:
```php
namespace App\Http;

use GuzzleHttp\Client;
use GuzzleHttp\Psr7\Request as GuzzleHttpRequest;
use Yoti\Http\RequestHandlerInterface;
use Yoti\Http\Request;
use Yoti\Http\Response;

/**
 * Handle HTTP requests.
 */
class RequestHandler implements RequestHandlerInterface
{
    /**
     * Execute HTTP request.
     *
     * @param Request $request
     *
     * @return \Yoti\Http\Response
     */
    public function execute(Request $request)
    {
        $body = null;
        if ($request->getPayload()) {
            $body = $request->getPayload()->getPayloadJSON();
        }

        $request = new GuzzleHttpRequest(
            $request->getMethod(),
            $request->getUrl(),
            $request->getHeaders(),
            $body
        );

        $client = new Client();
        $response = $client->send($request);

        return new Response(
            $response->getBody(),
            $response->getStatusCode()
        );
    }
}
```

Then pass an instance of this to `YotiClient::setRequestHandler()`

davidgrayston:
For reference, I've swapped cURL for Guzzle in the demo here: https://github.com/davidgrayston/yoti-php-sdk/commit/e9dfb06841b0d41c0df8974a09c383e0a3f2c239

davidgrayston:
@echarrod please review handling of slashes in 3ff6293b44e3976080fb8e39b1d09822960e45a3
- Base URL - always trim trailing slashes
- Endpoint - always has one leading slash (many leading slashes are trimmed to 1, and 1 slash is added if there are no leading slashes)

PR #92, opened 2019-10-10
SDK-1249: Re-organise demo

davidgrayston:
### Changed
- Moved public demo project files into public directory
- Selfie now uses `Yoti\Entity\Image::getBase64Content()` instead of writing to a local file
- Disabled `display_errors`
- Creating self-signed certificate in _Dockerfile_ (removed committed certificates)


examples/php.ini
```diff
@@ -0,0 +1,2 @@
+display_errors = off
```
echarrod: I think it might be better to display errors. If we're assuming that these are just _demo_ projects, and people won't just copy and paste them to use in their production environments, do we think it'd be more useful to show the error, rather than having to find it in the logs?

echarrod: In .NET and Go (maybe others), we're currently displaying the error

davidgrayston: We could hide them here, but display them in demo project built in pipeline?

davidgrayston: The errors will continue to show in the terminal when running the example

PR #93, opened 2019-10-11
SDK-1234: Attribute Issuance Details

davidgrayston:
#### Classes
-  `Yoti\Entity\AttributeIssuanceDetails`
-  `Yoti\Entity\AttributeDefinition`
-  `Yoti\Entity\ExtraData`
-  `Yoti\Exception\ExtraDataException`
- `Yoti\Util\ExtraData\DataEntryConverter`
- `Yoti\Util\ExtraData\ExtraDataConverter`
- `Yoti\Util\ExtraData\ThirdPartyAttributeConverter`
- `Yoti\Util\EncryptedData` to decrypt encrypted protobuf data

#### Methods
- `Yoti\ActivityDetails::getExtraData()`
- `Yoti\EntityReceipt::getExtraData()`
- `Yoti\Util\Validation::isArrayOfStrings()`

#### Generated `Sharepubapi` protobuf code:
  - `Sharepubapi\DataEntry`
  - `Sharepubapi\DataEntry\Type`
  - `Sharepubapi\Definition`
  - `Sharepubapi\ThirdPartyAttribute`
  - `Sharepubapi\ExtraData`
  - `Sharepubapi\IssuingAttributes`
  - `GPBMetadata\DataEntry`
  - `GPBMetadata\ThirdPartyAttribute`
  - `GPBMetadata\ExtraData`
  - `GPBMetadata\IssuingAttributes`

### Deprecated
- `\Yoti\Util\AttributeConverter::getEncryptedData`
- `\Yoti\Util\AttributeListConverter::convertToProtobufAttributeList `


src/Yoti/Entity/ExtraData.php
```diff
@@ -0,0 +1,37 @@
...
+    {
+        $attributeIssuanceDetails = array_filter(
+            $this->dataEntryList,
+            function ($dataEntry) {
+                return $dataEntry instanceof AttributeIssuanceDetails;
+            }
+        );
+
+        return reset($attributeIssuanceDetails) ?: null;
+    }
```
MrBurtyyy: When I spoke to Mike about this in Java, I think it was preferred to do the finding up front and only once.  This way we don't have to search for the `AttributeIssuanceDetails` multiple times if the user calls this method more than once?  What do you think?

src/Yoti/Entity/Receipt.php
```diff
@@ -104,6 +106,14 @@ public function parseAttribute($attributeName, $pem)
         );
     }
 
+    /**
+     * @return \Yoti\Entity\ExtraData
+     */
+    public function getExtraData()
+    {
+        return ExtraDataConverter::convertValue($this->getAttribute(self::ATTR_EXTRA_DATA_CONTENT));
+    }
```
MrBurtyyy: Again, isn't this susceptible to also being parsed multiple times?  I know it's unlikely for `getExtraData` to be called more than once, but I think it's better to maybe to do the parsing up front once?

src/Yoti/Entity/ExtraData.php
echarrod: Fair point, but can you envisage clients needing to call it more than once? It would probably be a code smell if they were calling the same method twice wouldn't it? (when nothing else has changed)

src/Yoti/Entity/Receipt.php
davidgrayston: Renamed to `parseExtraData` in bc018e8 (this is only called once when creating activity details)

src/Yoti/Entity/ExtraData.php
davidgrayston: 👍 Agree - optimised in bc018e8

davidgrayston:
#### Changes in a25d0ebc5afdc8833bd83fb1b64b569dbaa47625...ccb23212fe8833b8f395914cb48cabda1b4778aa
### Added
- `Yoti\Util\EncryptedData` to decrypt encrypted protobuf data

### Changed
- Token is now base64 encoded

### Deprected
- `\Yoti\Util\AttributeConverter::getEncryptedData`
- `\Yoti\Util\AttributeListConverter::convertToProtobufAttributeList `

davidgrayston:
I've updated `Sharepubapi` namespace in 317801e to be consistent with other generated protobuf code (we can move into Yoti namespace in next major release)

PR #97, opened 2019-11-07
SDK-1151: Allow multiple attributes with the same name

davidgrayston:
### Added
- `\Yoti\Entity\BaseProfile`
  - `::getAttributesByName()` - returns `\Yoti\Entity\Attribute[]`
  - `::getAttributesList()` - returns `\Yoti\Entity\Attribute[]`
     > Note: `::getAttributes()` - continues to return `\Yoti\Entity\Attribute[]` map keyed by attribute name.
- `\Yoti\Util\Profile\AttributeListConverter`
  - `::convertToYotiAttributesList()` - Converts Protobuf AttributeList to array of Yoti Attributes.

### Deprecated
- `\Yoti\Entity\BaseProfile::getAttributes()`
  > Replaced by `::getAttributesList()`
- `\Yoti\Entity\Profile::ATTR_AGE_VERIFICATIONS`
  > Age verifications are no longer appended to attributes list.
- `\Yoti\Util\Profile\AttributeListConverter::convertToYotiAttributesMap()`
  > Replaced by `::convertToYotiAttributesList()`

### Changed
- `\Yoti\Util\Profile\AttributeListConverter::convertToYotiAttributesMap()` will no longer include `null` items (attributes that could not be parsed) &dagger;
- Example project now supports multiple attributes
- Example project now supports new/unknown attributes

> &dagger; Although it's very unlikely that this class would be used outside the SDK, the behaviour has changed slightly in the following way:
> 
> Without `null` array item:
> ```php
> $some_array = ['key 1' => 'value 1'];
> 
> var_dump(isset($some_array['key 2']));
> // Output: bool(false)
> 
> var_dump($some_array['key 2'] === null);
> // Notice: Undefined index: key 2
> // Output: bool(true)
> 
> var_dump(array_key_exists('key 2', $some_array));
> // Output: bool(false)
> ```
> 
> With `null` array item:
> ```php
> $some_array = ['key 1' => 'value 1', 'key 2' => null];
> 
> var_dump(isset($some_array['key 2']));
> // Output: bool(false)
> 
> var_dump($some_array['key 2'] === null);
> // Output: bool(true)
> 
> var_dump(array_key_exists('key 2', $some_array));
> // Output: bool(true)
> ```
> 
> The current implementation requires that the value is always checked it is not `null`, so very unlikely to be a breaking change. There is a chance a notice would be generated if checked using `$some_array['key 2'] === null`

### DIFF
--- a/examples/public/dynamic-share.php
+++ b/examples/public/dynamic-share.php
@@ -12,22 +12,6 @@ use Yoti\ShareUrl\Policy\SourceConstraintBuilder;
 
 $yotiClient = new Yoti\YotiClient(YOTI_SDK_ID, YOTI_KEY_FILE_PATH);
 
-$drivingLicenseConstraint = (new ConstraintsBuilder())
-    ->withSourceConstraint(
-        (new SourceConstraintBuilder())
-            ->withDrivingLicence()
-            ->build()
-    )
-    ->build();
-
-$passportConstraint = (new ConstraintsBuilder())
-    ->withSourceConstraint(
-        (new SourceConstraintBuilder())
-            ->withPassport()
-            ->build()
-    )
-    ->build();
-
 $locationConstraint = (new LocationConstraintExtensionBuilder())
     ->withLatitude(50.8169)
     ->withLongitude(-0.1367)
@@ -36,9 +20,7 @@ $locationConstraint = (new LocationConstraintExtensionBuilder())
 
 $policy = (new DynamicPolicyBuilder())
     ->withFullName()
-    ->withGivenNames()
-    ->withDocumentDetails($drivingLicenseConstraint)
-    ->withDocumentDetails($passportConstraint)
+    ->withDocumentDetails()
     ->withDocumentImages()
     ->withAgeOver(18)
     ->withSelfie()
--- a/examples/public/dynamic-share.php
+++ b/examples/public/dynamic-share.php
@@ -6,9 +6,27 @@ require_once __DIR__ . '/../bootstrap.php';
 
 use Yoti\ShareUrl\DynamicScenarioBuilder;
 use Yoti\ShareUrl\Extension\LocationConstraintExtensionBuilder;
+use Yoti\ShareUrl\Policy\ConstraintsBuilder;
 use Yoti\ShareUrl\Policy\DynamicPolicyBuilder;
+use Yoti\ShareUrl\Policy\SourceConstraintBuilder;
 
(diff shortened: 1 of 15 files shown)


examples/public/dynamic-share.php
```diff
@@ -6,10 +6,28 @@
...
+    )
+    ->build();
+
+$passportConstraint = (new ConstraintsBuilder())
+    ->withSourceConstraint(
+        (new SourceConstraintBuilder())
+            ->withPassport()
+            ->build()
+    )
+    ->build();
```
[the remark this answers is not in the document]
echarrod: I think we decided we wouldn't, as it will only be a small subset of our client base who will want this, and it will put further restrictions on who can run the Dynamic Share demo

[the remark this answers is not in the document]
davidgrayston: Should we remove the source constraints example completely? If we don't want to demonstrate multiple attributes with different constraints, we could instead just have a single example of retrieving document details from a passport to simplify? I can also remove document images from the policy.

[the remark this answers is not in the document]
davidgrayston: Removed constraints in f39b11f

davidgrayston:
I've added additional coverage for wanted attributes with constraints and derivation in 3d3da16

examples/public/profile.php
```diff
@@ -1,5 +1,119 @@
...
+            <td>Result</td>
+            <td><?php echo htmlspecialchars($ageVerification->getResult()); ?></td>
+        </tr>
+    </table>
+    <?php
+}
+
+
+/**
+ * Format Structure Postal Address.
```
echarrod:
```suggestion
 * Format Structured Postal Address.
```

src/Yoti/Entity/Profile.php
```diff
@@ -25,10 +26,16 @@ class Profile extends BaseProfile
     const ATTR_DOCUMENT_IMAGES = 'document_images';
     const ATTR_STRUCTURED_POSTAL_ADDRESS = 'structured_postal_address';
 
+    /** @deprecated 3.0.0 No longer used to store age verifcations. */
```
echarrod:
```suggestion
    /** @deprecated 3.0.0 No longer used to store age verifications. */
```

examples/public/profile.php
davidgrayston: Corrected in 85db368

src/Yoti/Entity/Profile.php
davidgrayston: Corrected in 85db368

davidgrayston:
Superseded by #138

PR #124, opened 2020-01-15
SDK-1344: Deprecate old namespaces classes

davidgrayston:
### Added
-  Aliasing old namespaces to new with deprecation message


davidgrayston:
Replaced by #125

PR #127, opened 2020-01-16
SDK-1346: Add type hints and return types

davidgrayston:
### Changed
- Enabled [strict types](https://www.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration.strict) in all PHP files
- Methods previously expecting `$pem` (string), is now `PemFile $pemFile`
- `Yoti\Http\Exception\NetworkException` / `Yoti\Http\Exception\RequestException` no longer has an exception code _(this follows HTTPlug and Guzzle implementations)_

### Added
- Scalar type hints
- Method return types
- `phpstan/phpstan-strict-rules`
  - Added `phpstan analyse` to `composer lint` command
  - Checks `src` and `sandbox/src`
- `Yoti\Util\DateTime::createFromFormat()` that wraps `\DateTime::createFromFormat()`, setting UTC timezone and throws on error - this is to reduce duplication of _DateTime_ conversion code
- `Yoti\Util\Json::encode()` that wraps `\json_encode()`, but throws on error to allow consistent `string` return type

> Note: [see valid types here](https://www.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration)

### Removed
- `AttributeListConverter::convertToProtobufAttributeList()` _(deprecated)_

### Fixed
- Exception message tests incorrectly converted in #121 now use `::expectExceptionMessage()`
- `Receipt::decryptAttribute()` was assuming the receipt attribute/item was always available, now:
  - Returns `null` when not available
  - Returns `string` when available and decrypts
  - Throws `EncryptedDataException` when available but fails to decrypt
- `ExtraDataConverter::convertValue()` now returns empty `ExtraData` early when provided value is `null`
- Anchors without a value have empty string value `""` _(was `null`)_


tests/Profile/ServiceTest.php
```diff
@@ -133,6 +134,29 @@ public function testInvalidConnectToken()
...
+    /**
+     * Test invalid Token
+     *
+     * @covers ::getActivityDetails
+     * @covers ::decryptConnectToken
+     */
+    public function testWrongPemFile()
+    {
+        $this->expectException(\Yoti\Exception\ActivityDetailsException::class);
+        $this->expectExceptionMessage('Could not decrypt connect token');
```
echarrod: I know you're just testing for this exception message here, and you haven't added it, but "Connect" isn't something which is known outside of Yoti (I know we have `YOTI_CONNECT_TOKEN` and `INVALID_YOTI_CONNECT_TOKEN` too).
So I don't think this is a very helpful error message to anyone outside of Yoti. 

Could we instead refer to it as the one time use token, as we do in the README and documentation?

davidgrayston: Re-worded in e804d5f (I've left the test data as is)

PR #138, opened 2020-01-31
SDK-1151: Allow multiple attributes with the same name

davidgrayston:
> Ported from #97 

### Changed
- `Yoti\Profile\BaseProfile::getAttributesByName()` is now public
- `Yoti\ShareUrl\Policy\DynamicPolicyBuilder::withWantedAttribute()` now allows the same wanted attribute with different constraints


echarrod:
Looking at https://github.com/getyoti/yoti-php-sdk/pull/97, it looks like the examples used to be in the `public` folder. Could you copy the changes in this PR which are in `public` to the current file locations, and we can see if there are any changes?

LuborRod:
> 

I have already checked everything, the same logic. It will not be same code, because of in new examples we use blade from laravel for view.

tests/Profile/BaseProfileTest.php
```diff
@@ -69,7 +69,6 @@ public function setup(): void
      * @covers ::__construct
      * @covers ::getProfileAttribute
      * @covers ::setAttributesMap
-     * @covers ::getAttributesByName
```
echarrod, line 72: Should this line have been removed?

echarrod (commented):
I think we still have some of the changes in https://github.com/getyoti/yoti-php-sdk/pull/97 that we need to have in this PR too: 
- src/Yoti/Util/Profile/AttributeListConverter.php (convertToYotiAttributesList)
- other changes in src/Yoti/Entity/BaseProfile.php (attributesMap, attributesList)
- the accompanying tests for these

tests/Profile/BaseProfileTest.php
LuborRod: 1. We have this method here - Yoti\Profile\Util\Attribute\AttributeListConverter and tests in Yoti\Test\Profile\Util\Attribute\AttributeListConverterTest.

2.We have these methods in another classes also, instead set we use construct method and convert everything before call constructor in class above.

3. All these methods are covered by tests, we have 100% )

4. Should be removed - David created special test for this method on 95 line.

echarrod (approved):
I see those changes I mentioned are already in development, so all good :+1:

PR #141, opened 2020-02-04
SDK-1303: Introduce Doc Scan client

MrBurtyyy:
## Added

* New Doc Scan client used for signed request operations
  * `Yoti\Docs\DocScanClient`
    * `::createSession()`
    * `::getSession()`
    * `::deleteSession()`
    * `::getMedia()`
    * `::deleteMedia()`
  * `Yoti\Docs\Exception\DocScanException`

* New `MediaFactory` class used to create Media objects based on `Content-Type` header
  * `Yoti\Media\MediaFactory`
    * `::fromMimeType()`

## Changed

* `Yoti\Test\TestCase::captureExpectedLogs()` has been updated to use `realpath` instead of using a relative path.  This should make it more compatible with Windows/Unix systems.
* Log files are also deleted after each test run to stop the temporary logs directory from filling up.


src/Docs/DocScanClient.php
```diff
@@ -0,0 +1,120 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yoti\Docs;
```
davidgrayston: Should we make the namespace `Yoti\DocScan`?

src/Docs/Session/Retrieve/TaskResponse.php
```diff
@@ -0,0 +1,133 @@
...
+     */
+    public function getState(): ?string
+    {
+        return $this->state;
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getCreated(): ?string
```
emmas-yoti: Do we want to call this created_date instead? To make it a little more self-documenting?

davidgrayston: I think this will likely become a `DateTime` in PHP - we can just use `Yoti\Util\DateTime::stringToDateTime()`

echarrod: > Do we want to call this created_date instead? To make it a little more self-documenting?

I think Alex replied about this in the other PR, but we're generally just reflecting the JSON in these function names, so I think we were reluctant to deviate from them. 

As David mentions, it will give more clarity anyway when the type is a native DateTime object.

src/Docs/Session/Create/NotificationConfigBuilder.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public function forResourceUpdate(): self
+    {
+        return $this->withTopic(self::RESOURCE_UPDATE);
+    }
+
+    public function withTopic(string $topic): self
+    {
+        if (!in_array($topic, $this->topics, true)) {
```
davidgrayston: We don't have this check in other SDKs - do you think this is something we should do across the board?

cc @echarrod

src/Docs/Session/Create/Check/RequestedLivenessCheckConfig.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yoti\Docs\Session\Create\Check;
+
+class RequestedLivenessCheckConfig implements RequestedCheckConfigInterface
```
davidgrayston: To match Java (perhaps not critical)
```suggestion
class RequestedLivenessConfig implements RequestedCheckConfigInterface
```

src/Docs/Session/Retrieve/LivenessResourceResponse.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Yoti\Docs\Session\Retrieve;
+
+abstract class LivenessResourceResponse extends ResourceResponse
+{
+
+    /**
+     * @return string
+     */
+    abstract public function getLivenessType(): string;
```
davidgrayston: This method isn't needed as each liveness resource has it's own concrete implementation

src/Docs/Session/Retrieve/ZoomLivenessResourceResponse.php
```diff
@@ -0,0 +1,75 @@
...
+        foreach ($frames as $frame) {
+            $parsedFrames[] = new FrameResponse($frame);
+        }
+        return $parsedFrames;
+    }
+
+    /**
+     * @return string
+     */
+    public function getLivenessType(): string
```
davidgrayston: We should be able to remove this method

src/Docs/Session/Create/NotificationConfigBuilder.php
echarrod: Do you mean not adding the topic if it's already there? Seems sensible

echarrod: And we'd also obviously want a unit test to confirm that behaviour in each SDK if we do that

src/Docs/Session/Create/NotificationConfig.php
```diff
@@ -0,0 +1,75 @@
...
+     * NotificationConfig constructor.
+     * @param string|null $authToken
+     * @param string|null $endpoint
+     * @param array<string> $topics
+     */
+    public function __construct(?string $authToken, ?string $endpoint, array $topics = [])
+    {
+        $this->authToken = $authToken;
+        $this->endpoint = $endpoint;
+        $this->topics = $topics;
```
davidgrayston: You could do `array_unique($topics)` here instead of de-duping in `withTopic()`

src/Docs/Session/Create/SdkConfig.php
```diff
@@ -0,0 +1,152 @@
...
+            'preset_issuing_country' => $this->getPresetIssuingCountry(),
+            'success_url' => $this->getSuccessUrl(),
+            'error_url' => $this->getErrorUrl(),
+        ];
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getAllowedCaptureMethods(): ?string
```
davidgrayston: We could cut down the size of the `Yoti\Docs\Session\Create\*` classes by removing the public getters, as they only need to serialise to JSON _(it would also cut down the number of unit tests required to test these classes)_. This is only a suggestion, happy to keep them if we think they'll be useful in future.

echarrod: If we can't think of a use case for exposing them, I think we should remove them. We can always add the getters at a later stage

src/DocScan/Session/Retrieve/BreakdownResponse.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $this->subCheck = $breakdown['sub_check'] ?? null;
+        $this->result = $breakdown['result'] ?? null;
+
+        if (isset($breakdown['details'])) {
+            $details = [];
+            foreach ($breakdown['details'] as $detail) {
+                $details[] = new DetailsResponse($detail['name'], $detail['value']);
+            }
+            $this->details = $details;
```
davidgrayston:
```suggestion
            foreach ($breakdown['details'] as $detail) {
                $this->details[] = new DetailsResponse($detail['name'], $detail['value']);
            }
```

src/DocScan/Session/Retrieve/CheckResponse.php
```diff
@@ -0,0 +1,142 @@
...
+     * @throws DateTimeException
+     */
+    public function __construct(array $checkData)
+    {
+        $this->id = $checkData['id'] ?? null;
+        $this->state = $checkData['state'] ?? null;
+        $this->resourcesUsed = $checkData['resources_used'] ?? [];
+
+        if (isset($checkData['generated_media'])) {
+            $this->generatedMedia = $this->parseGeneratedMedia($checkData['generated_media']);
```
davidgrayston: You could replace the private method with this _(just a suggestion, fine to leave this as is)_

```suggestion
        foreach ($checkData['generated_media'] as $generatedMedia) {
            $this->generatedMedia[] = new GeneratedMedia($generatedMedia);
        }
```

src/DocScan/Session/Retrieve/GetSessionResult.php
```diff
@@ -0,0 +1,196 @@
...
+        $this->userTrackingId = $sessionData['user_tracking_id'] ?? null;
+        $this->clientSessionToken = $sessionData['client_session_token'] ?? null;
+        $this->clientSessionTokenTtl = $sessionData['client_session_token_ttl'] ?? null;
+
+        if (isset($sessionData['checks'])) {
+            $parsed = [];
+            foreach ($sessionData['checks'] as $check) {
+                $parsed[] = $this->createCheckFromArray($check);
+            }
+            $this->checks = $parsed;
```
davidgrayston:
```suggestion
            foreach ($sessionData['checks'] as $check) {
                $this->checks[] = $this->createCheckFromArray($check);
            }
```

src/DocScan/Session/Retrieve/IdDocumentResourceResponse.php
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public function __construct(array $idDocument)
+    {
+        parent::__construct($idDocument);
+
+        $this->documentType = $idDocument['document_type'] ?? null;
+        $this->issuingCountry = $idDocument['issuing_country'] ?? null;
+
+        if (isset($idDocument['pages'])) {
+            $this->pages = $this->parsePages($idDocument['pages']);
```
davidgrayston: Could replace private method with:

```suggestion
            foreach ($idDocument['pages'] as $page) {
                $this->pages[] = new PageResponse($page);
            }
```

src/DocScan/Session/Retrieve/ReportResponse.php
```diff
@@ -0,0 +1,63 @@
...
+     * @param array<string, mixed> $reportData
+     */
+    public function __construct(array $reportData)
+    {
+        if (isset($reportData['recommendation'])) {
+            $this->recommendation = new RecommendationResponse($reportData['recommendation']);
+        }
+
+        if (isset($reportData['breakdown'])) {
+            $this->breakdown = $this->parseBreakdowns($reportData['breakdown']);
```
davidgrayston: Could replace private method with:
```suggestion
            foreach ($reportData['breakdown'] as $breakdown) {
                $this->breakdown[] = new BreakdownResponse($breakdown);
            }
```

src/DocScan/Session/Retrieve/BreakdownResponse.php
MrBurtyyy: Fixed in 93f3254

src/DocScan/Session/Retrieve/CheckResponse.php
MrBurtyyy: Fixed in 93f3254

src/DocScan/Session/Retrieve/GetSessionResult.php
MrBurtyyy: Fixed in 93f3254

src/DocScan/Session/Retrieve/IdDocumentResourceResponse.php
MrBurtyyy: Fixed in 93f3254

src/DocScan/Session/Retrieve/ReportResponse.php
MrBurtyyy: Fixed in 93f3254

src/DocScan/Session/Retrieve/GetSessionResult.php
davidgrayston:
```suggestion
            foreach ($sessionData['checks'] as $check) {
                $this->checks[] = $this->createCheckFromArray($check);
            }
```

src/DocScan/Session/Retrieve/ResourceResponse.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /**
+     * @return TaskResponse[]
+     */
+    public function getTasks(): array
+    {
+        return $this->tasks;
+    }
+}
```
davidgrayston: I think we also want `::getTextExtractionTasks()` here in case we introduce new tasks

src/DocScan/Session/Retrieve/TaskResponse.php
```diff
@@ -0,0 +1,142 @@
...
+        return $this->lastUpdated;
+    }
+
+    /**
+     * @return GeneratedCheckResponse[]
+     */
+    public function getGeneratedChecks(): array
+    {
+        return $this->generatedChecks;
+    }
```
davidgrayston: We will also need `::getGeneratedTextDataChecks()` in case we introduce additional checks

MrBurtyyy: Fixed in 55a0dfa

src/DocScan/Session/Retrieve/ResourceResponse.php
MrBurtyyy: Fixed in 55a0dfa

src/DocScan/Session/Retrieve/GetSessionResult.php
MrBurtyyy: Fixed in 55a0dfa

src/DocScan/Session/Retrieve/ResourceResponse.php
davidgrayston: I can see this fixed in 77ce40a

echarrod:
> Nice 👍
> 
> @echarrod do we want to add a demo in a separate PR?

Yes I think to try to minimise the size of this one, we should do it in a separate PR

davidgrayston:
@echarrod I've rebased this ready to merge - can you take a looks at changes in 649e396?

PR #144, opened 2020-02-11
SDK-1446: Document Details Validation

davidgrayston:
### Fixed
- Allow any character in DocumentDetails value

### DIFF
--- a/src/Profile/Attribute/DocumentDetails.php
+++ b/src/Profile/Attribute/DocumentDetails.php
@@ -14,7 +14,6 @@ class DocumentDetails
      * The values of the Document Details are in the format and order as defined in this pattern
      * e.g PASS_CARD GBR 22719564893 - CITIZENCARD, the last two are optionals
      */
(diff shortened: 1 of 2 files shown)


src/Profile/Attribute/DocumentDetails.php
```diff
@@ -117,13 +115,33 @@ public function getIssuingAuthority(): ?string
...
+     */
+    private function validateParsedValues(array $parsedValues): void
+    {
+        if (count($parsedValues) < 3) {
+            throw new AttributeException('Invalid value for DocumentDetails');
+        }
+
+        foreach ($parsedValues as $parsedValue) {
+            if (strlen($parsedValue) === 0) {
+                throw new AttributeException('DocumentDetails cannot have empty values');
```
echarrod: In this scenario I don't know if this would make much sense to have "empty values", as this could just be an extra space in a OCR-read value. 
I've just put "Invalid Document Details":
```suggestion
                throw new AttributeException('Invalid DocumentDetails value');
```
That might make more sense, what do you think?

davidgrayston: Rebased and fixed in ef55028

PR #153, opened 2020-03-17
SDK-1303: Doc Scan Demo

davidgrayston:
> Templates copied from https://github.com/getyoti/yoti-node-sdk/pull/142

### Added
- Doc Scan demo (Laravel app)
  > Note: Many of the files are defaults from https://github.com/laravel/laravel - we may be able remove some of the config files, but there's no harm having them


davidgrayston:
@echarrod can you check that changes in acf9a4f work for you?

PR #156, opened 2020-03-30
SDK-1303: Add link to Doc Scan example

davidgrayston:
> Correction to #153

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,12 +33,10 @@ How to integrate with Yoti's AML (Anti Money Laundering) service
 
 8) [How to run the examples](#how-to-run-the-examples)
 
-9) [Running the tests](#running-the-tests)
-
-10) [API Coverage](#api-coverage) -
+9) [API Coverage](#api-coverage) -
 Attributes defined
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,12 +33,10 @@ How to integrate with Yoti's AML (Anti Money Laundering) service
 
 8) [How to run the examples](#how-to-run-the-examples)
 
-9) [Running the tests](#running-the-tests)
-
-10) [API Coverage](#api-coverage) -
+9) [API Coverage](#api-coverage) -
 Attributes defined
 
-11) [Support](#support) -
```
echarrod: Not essential here, but next time we update this we can actually just change the numbers to all be `1)`, and markdown calculates the numbers for you. This prevents any potential forgetting to update the numbers. See "raw" view of [.NET README](https://github.com/getyoti/yoti-dotnet-sdk/blob/master/README.md)

davidgrayston: We could make this an unordered list - not sure if numbered list makes sense here

PR #158, opened 2020-04-08
SDK-1501: API URL environment variable

davidgrayston:
### Added
- Default API URLs can now be overridden using environment variables:
  - `YOTI_API_URL` to override the default API URL
  - `YOTI_DOC_SCAN_API_URL` to override the default Doc Scan API URL

### DIFF
--- a/tests/DocScan/DocScanClientTest.php
+++ b/tests/DocScan/DocScanClientTest.php
@@ -36,6 +36,107 @@ class DocScanClientTest extends TestCase
     /**
      * @test
      * @covers ::__construct
+     */
+    public function testDefaultApiUrl()
+    {
+        $response = $this->createMock(ResponseInterface::class);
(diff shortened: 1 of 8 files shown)


tests/DocScan/DocScanClientTest.php
```diff
@@ -33,6 +33,107 @@ public function shouldThrowExceptionForEmptySdkId()
         new DocScanClient('', TestData::PEM_FILE);
     }
 
+    /**
+     * @test
+     * @covers ::__construct
+     */
+    public function testDefaultApiUrl()
+    {
+        $response = $this->createMock(ResponseInterface::class);
```
echarrod, line 42: Do we need to make sure that the env variable isn't set here too?

echarrod: Never mind, saw the removal in `TestCase.php` after I left that comment!

PR #159, opened 2020-04-17
SDK-1463: Multiple Documents and Filters

davidgrayston:
### Added
- `DocScanClient::getSupportedDocuments()` returns list of supported documents for each country
- `SessionSpecificationBuilder::withRequiredDocument()`
  - `RequiredIdDocumentBuilder` builds a required identity document
    - `DocumentRestrictionsFilterBuilder` builds document restrictions filter for required documents
      - `DocumentRestrictionBuilder` builds document restriction for document restrictions filter
    - `OrthogonalRestrictionsFilterBuilder` builds orthogonal restrictions filter for required documents


src/DocScan/Session/Create/Filters/RequiredDocument.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /**
+     * @return \stdClass
+     */
+    public function jsonSerialize(): \stdClass
+    {
+        return (object) Json::withoutNullValues([
+            'type' => $this->type,
+            'filter' => $this->filter,
```
MrBurtyyy, line 38: I've spoken with Mike, and we're going to move the filter onto the RequiredIdentityDocument interface.  This is due to the fact that they know all RequiredDocuments will have at least a `type`, however they don't know if they will always have a filter (e.g. for utility bills etc).

davidgrayston: Fixed in 2059a79

davidgrayston:
Changes made to make country codes and document types nullable in 83f6c52

PR #188, opened 2020-06-27
SDK-1628: Protobuf C Extension

davidgrayston:
- [x] Merge #115 and rebase this PR

### Added
- Travis job to test with [protobuf C extension](https://github.com/protocolbuffers/protobuf/tree/master/php) ([pecl](https://pecl.php.net/package/protobuf)) installed - integrators should install this extension for performance improvements

### Changed
- No longer mocking protobuf classes _(PHPUnit cannot mock protobuf classes with extension enabled)_
- Attribute converter now skips attributes with name that is `null` or empty string `""`
  > Note: it's not possible to create a protobuf attribute with `null` name without mocking, so we could remove this condition. When the name is not set, it will default to an empty string `""`, so skipping when the name is `null` or `""` seemed sensible.
- Exceptions messages are shorter with the extension enabled, so tests have been updated to check for the shorter version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,9 @@ Please feel free to reach out
 * PHP ^7.1
 * CURL PHP extension (must support TLSv1.2)
 
+### Recommended (optional)
+- [Protobuf C extension](https://github.com/protocolbuffers/protobuf/tree/master/php) (PHP package will be used by default)
+
 ## Enabling the SDK
 
 To import the Yoti SDK inside your project, you can use your favourite dependency management system.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -30,6 +30,9 @@ Please feel free to reach out
 * PHP ^7.1
 * CURL PHP extension (must support TLSv1.2)
 
+### Extras
+- [Protobuf C extension](https://github.com/protocolbuffers/protobuf/tree/master/php) (PHP package will be used by default)
+
```
davidgrayston, line 35: @jlmartyres does this read ok?

PR #189, opened 2020-07-06
SDK-1635: Chip Data

davidgrayston:
### Added
- Doc Scan chip data support


src/DocScan/Session/Create/Task/RequestedTextExtractionTaskBuilder.php
```diff
@@ -12,38 +12,89 @@ class RequestedTextExtractionTaskBuilder
...
+     * @return $this
+     */
     public function withManualCheckFallback(): self
     {
         return $this->withManualCheck(self::FALLBACK);
     }
 
+    /**
+     * @return $this
+     */
```
echarrod, line 57: Is there a step we can add to pre-commit or Travis to ensure functions have type hints?

davidgrayston: We should be able to enforce this, but they're not critical as we now have return types on methods. We already check that these are correct when present, but they're not mandatory.

PR #208, opened 2020-09-08
SDK-1753: Remove Face Map from demo

davidgrayston:
### Removed
- Face Map is removed from demo


echarrod (approved):
You did it before I had chance to create the sub-ticket 😄 

Doesn't matter about the commit title, but if you could update the PR title to SDK-1753, that would be great 👍

PR #212, opened 2020-09-21
IN-3117: php Example QR

nikhilPank:


echarrod (approved):

davidgrayston (changes requested):
Do we need to make this change in the main share example too?
https://github.com/getyoti/yoti-php-sdk/blob/development/examples/profile/app/Http/Controllers/ShareController.php#L21

davidgrayston (approved):

PR #228, opened 2021-06-07
SDK-1829: Add Third Party Identity Check

LuborRod:
Add third-party-identity-check


echarrod (approved):
Looks good, one the example has been tested, it will be good to do an interactive rebase and `fixup` the duplicate commits :+1:

PR #229, opened 2021-06-16
SDK-1840:Add auth type to notification config

LuborRod:
Add auth type to notification config

### DIFF
--- a/src/DocScan/Session/Create/NotificationConfig.php
+++ b/src/DocScan/Session/Create/NotificationConfig.php
@@ -32,11 +32,11 @@ class NotificationConfig implements \JsonSerializable
     /**
      * NotificationConfig constructor.
      * @param string|null $authToken
-     * @param string|null $authType
      * @param string|null $endpoint
      * @param array<string> $topics
+     * @param string|null $authType
      */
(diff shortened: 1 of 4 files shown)


src/DocScan/Session/Create/NotificationConfig.php
```diff
@@ -27,12 +32,14 @@ class NotificationConfig implements \JsonSerializable
     /**
      * NotificationConfig constructor.
      * @param string|null $authToken
+     * @param string|null $authType
      * @param string|null $endpoint
      * @param array<string> $topics
      */
-    public function __construct(?string $authToken, ?string $endpoint, array $topics = [])
+    public function __construct(?string $authToken, ?string $authType, ?string $endpoint, array $topics = [])
```
echarrod, line 39: I know no-one _should_ be calling this constructor, since we want them to use the builder. But if they are, will the addition of an argument (which isn't the last one) cause existing calls to this method to use the wrong parameters?

echarrod (approved):
Nice, looks good now :+1:

bucky-boy (approved):
Appears to be ok from what I see, not being a PHP dev.

Please ensure it gets QA'd properly before it goes out.

PR #230, opened 2021-06-18
SDK-1755:Add http exception response

LuborRod:
Add http exception response


src/Http/Response.php
```diff
@@ -0,0 +1,46 @@
...
+    /** HTTP codes */
+    private const BAD_REQUEST_CODE = 400;
+    private const UNAUTHORIZED_CODE = 401;
+    private const INTERNAL_SERVER_CODE = 500;
+
+    /**
+     * @param ResponseInterface $response
+     * @throws DocScanException
+     */
+    public static function createYotiExceptionFromStatusCode(ResponseInterface $response): void
```
echarrod, line 27: Doesn't look like this is used anywhere but the tests. We need to include the HTTP response on any exception that we return to clients

src/Aml/Service.php
```diff
@@ -47,7 +48,7 @@ public function __construct(string $sdkId, PemFile $pemFile, Config $config)
      *
      * @return \Yoti\Aml\Result
      *
-     * @throws \Yoti\Exception\AmlException
+     * @throws YotiException
```
echarrod, line 51: Not sure I understand this change, we're doing `throw new AmlException` below, so does that now mean this should still be annotated with `@throws AmleException`?

LuborRod: Fixed.

echarrod (approved):

PR #234, opened 2021-06-29
Sdk-1902:Doc scan watchlist screening check

LuborRod:
## Added

* Support for WATCHLIST_SCREENING check in Doc Scan


phpunit.xml
```diff
@@ -18,6 +18,7 @@
             <directory suffix=".php">src</directory>
             <exclude>
                 <directory>src/Protobuf</directory>
+                <directory>src/DocScan/Session/Retrieve/Contracts</directory>
```
echarrod, line 21: Generally we take the approach of only excluding generated files, so can we remove this line? 
```suggestion
```

echarrod (changes requested):
Just a small PHPUnit exclusion change, think it looks good otherwise :+1:

phpunit.xml
LuborRod: Yes

echarrod (approved):
:+1:

echarrod:
Merging with failing quality gate as coverage will be added in a separate PR

PR #235, opened 2021-07-07
 SDK-1913 add support for watchlist advanced ca check

LuborRod:
Add support for WATCHLIST_ADVANCED_CA check

### DIFF
--- a/examples/doc-scan/app/Http/Controllers/HomeController.php
+++ b/examples/doc-scan/app/Http/Controllers/HomeController.php
@@ -10,6 +10,7 @@ use Yoti\DocScan\Session\Create\Check\RequestedFaceMatchCheckBuilder;
 use Yoti\DocScan\Session\Create\Check\RequestedIdDocumentComparisonCheckBuilder;
 use Yoti\DocScan\Session\Create\Check\RequestedLivenessCheckBuilder;
 use Yoti\DocScan\Session\Create\Check\RequestedThirdPartyIdentityCheckBuilder;
+use Yoti\DocScan\Session\Create\Check\RequestedWatchlistAdvancedCaCheckBuilder;
 use Yoti\DocScan\Session\Create\Check\RequestedWatchlistScreeningCheckBuilder;
 use Yoti\DocScan\Session\Create\Check\RequestedWatchlistScreeningConfigBuilder;
 use Yoti\DocScan\Session\Create\Filters\Orthogonal\OrthogonalRestrictionsFilterBuilder;
@@ -59,6 +60,10 @@ class HomeController extends BaseController
             ->withRequestedCheck(
                 (new RequestedWatchlistScreeningCheckBuilder())
                     ->withConfig($watchScreeningConfig)
+                    ->build()
+            )
+            ->withRequestedCheck(
+                (new RequestedWatchlistAdvancedCaCheckBuilder())
                     ->build()
             )
             ->withRequestedTask(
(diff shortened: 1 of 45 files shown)


echarrod (approved):

src/DocScan/Session/Create/Check/RequestedCustomAccountWatchlistAdvancedCaConfig.php
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * RequestedCustomAccountWatchlistAdvancedCaConfig constructor.
+     * @param bool $removeDeceased
+     * @param bool $shareUrl
+     * @param RequestedCaSources $sources
+     * @param RequestedCaMatchingStrategy $matchingStrategy
+     * @param string $apiKey
+     * @param bool $monitoring
+     * @param string[] $tags
```
laurent-yoti, line 41: @LuborRod A question for you, I have been basing the Node SDK mostly on your work, and I was wondering about this `tags`, as the documentation says it should be a map and it looks like you have a list of string (but I am not aware of Php, maybe there is something I am missing).
Documentation:

```"tags": { "someKey": "someValue" }, // optional map, see below```
```
Tags
The map of tags is dependent on the api_key - we have no way of knowing what should or should not be allowed.  We will only accept tags if they are giving us their api_key.  The map is optional, we will allow an empty map, but we will not allow…
null map
null keys or values
```

Comment (LuborRod) [reply src/DocScan/Session/Create/Check/RequestedCustomAccountWatchlistAdvancedCaConfig.php:41]: @laurent-yoti In PHP, arrays serve as all types of containers. They can be used as maps, lists and sets.

Comment (laurent-yoti) [reply src/DocScan/Session/Create/Check/RequestedCustomAccountWatchlistAdvancedCaConfig.php:41]: 👍

File: src/DocScan/Session/Retrieve/Contracts/WatchlistAdvancedCaSearchConfigResponse.php [hunk 2]
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * @var bool
+     */
+    private $shareUrl;
+
+    /**
+     * @var CaSourcesResponse
+     */
+    private $sources;
```
bucky-boy, line 27: Not knowing php, I don't immediately see how someone would retrieve _sources_ or _matching_strategy_...?

examples/doc-scan/app/Http/Controllers/HomeController.php
```diff
@@ -61,6 +62,10 @@ public function show(Request $request, DocScanClient $client)
                     ->withConfig($watchScreeningConfig)
                     ->build()
             )
+            ->withRequestedCheck(
+                (new RequestedWatchlistAdvancedCaCheckBuilder())
+                    ->build()
```
bucky-boy, line 67: no config on this one?

bucky-boy (approved):
Looks very close to the Java SDK.  Neatly done, nice.

The risk with releasing this now is that it cannot (yet) be tested against the sandbox.  Sandbox support is coming, but will be at least another couple of weeks.

If you want to press ahead with the release please ensure it's tested carefully.  You'd need to double check it works for both forms of _sources_ and _matching_strategy_, and also ensure you can retrieve the generated _raw_reults_ etc buried inside the report.

Also, compare the demo with the Java SDK version.  Pawel QA'd that one closer to the time we released it so it should be a very reliable guide to your expectations.

src/DocScan/Session/Retrieve/Contracts/WatchlistAdvancedCaSearchConfigResponse.php
LuborRod: Fixed

examples/doc-scan/app/Http/Controllers/HomeController.php
LuborRod: Fixed

PR #242, opened 2021-11-12
SDK-2000 Add support for mobile handoff

LuborRod:


tests/DocScan/Session/Create/SdkConfigBuilderTest.php
```diff
@@ -123,6 +128,7 @@ public function shouldProduceTheCorrectJsonString()
             'success_url' => self::SOME_SUCCESS_URL,
             'error_url' => self::SOME_ERROR_URL,
             'privacy_policy_url' => self::SOME_PRIVACY_POLICY_URL,
+            'allow_handoff' => true,
```
echarrod, line 131: Please could we also add a small test that checks that when this value isn't set, it's value is `null`?

echarrod (approved):
:ok_hand:

PR #251, opened 2021-12-29
Sdk 1956 Allow Relying Business to fetch IBV Instructions

LuborRod:


src/DocScan/DocScanClient.php
```diff
@@ -184,4 +185,14 @@ public function putIbvInstructions(string $sessionId, Instructions $instructions
     {
         $this->docScanService->putIbvInstructions($sessionId, $instructions);
     }
+
+    /**
+     * @param string $sessionId
```
echarrod, line 190:
```suggestion
     * Fetches the instructions PDF associated with an In-Branch Verification session.
     *
     * @param string $sessionId
```

echarrod (approved):
Looks good, just a minor point about adding a comment. If we could do this on the places the clients are most likely going to inspect from their code, that would be great :+1:

PR #254, opened 2022-01-05
Release/3.8.0 

LuborRod:
# Changes

## Added

- Support for non-latin documents at session creation
- Support for non-latin documents when fetching session configuration
- Support for non-latin documents when fetching supported documents


src/DocScan/DocScanClient.php
```diff
@@ -129,13 +129,13 @@ public function deleteMediaContent(string $sessionId, string $mediaId): void
     }
 
     /**
-     * Gets a list of supported documents.
```
echarrod, line 132: Ah, presume this comment wasn't intentionally removed?

echarrod (commented):
Looks good, just a function comment I noticed has been removed

jlmartyres (approved):

src/DocScan/DocScanClient.php
LuborRod: I see, I will add it. Thx.

PR #263, opened 2022-02-02
Return null when we receive 204 from getMediaContent

LuborRod:


src/DocScan/Service.php
```diff
@@ -28,6 +28,9 @@
 
 class Service
 {
+    /** @const int */
+    private const HTTP_STATUS_NO_CONTENT = 204;
```
echarrod, line 32: Do we have other similar constants defined elsewhere? just wondering if this should live somewhere else

echarrod (approved):

src/DocScan/Service.php
LuborRod: For now - not)

PR #269, opened 2022-02-21
Upgraded Guzzle PSR7 to version 2.1.0 and moved psr/log to require-dev

tyloo:
Title says it all. This allow the use of the latest versionsof Guzzle PSR7 and the fact to move psr/log to require-dev allow us to not have conflicts while being only used for tests.


tyloo:
Any updates on this guys ? Should be safe :)

echarrod:
Thanks for the PR @tyloo! This will be in our upcoming [v4.0.0 release](https://github.com/getyoti/yoti-php-sdk/pull/281)

PR #275, opened 2022-03-24
SDK-2075 SDK-2085 Add subject and identity profile requirements

LuborRod:


tests/ShareUrl/DynamicScenarioBuilderTest.php
```diff
@@ -67,6 +74,7 @@ public function testBuild()
                 'wanted_auth_types' => [],
                 'wanted_remember_me' => false,
                 'wanted_remember_me_optional' => false,
+                'identity_profile_requirements' => null
```
echarrod, line 77: I think it's best to be explicit here, and include this null value (like you've done), rather than omitting it. @BenSmithYoti I think this matches the .NET implementation, right?

echarrod (approved):
All good from my side, would be good to get @BenSmithYoti's approval before merging :+1:

BenSmithYoti (approved):
Responded to Ed's comment. All looks good to me. Thanks Rodion. Perfect!

tests/ShareUrl/DynamicScenarioBuilderTest.php
BenSmithYoti: Another e.g. is `issuing_authority_sub_check ` on `RequestedDocumentAuthenticityConfig`  

In short I believe the implementation in this PR will send up the following JSON: `"identity_profile_requirements" : null,` and I believe this will be handled fine by the API (same behaviour as if identity_profile_requirements was excluded from the JSON).

PR #276, opened 2022-03-31
SDK-2080 Add identity profile report

LuborRod:


echarrod (approved):
Looks good. I realised we will need to add this attribute to be displayed in the Example project too, but that can be done in a separate PR, doesn't have to be in this one

PR #284, opened 2022-05-23
SDK-2143: Add error code and description in failure receipt message

LuborRod:


examples/profile/app/Http/Controllers/DbsCheckController.php
```diff
@@ -16,13 +16,13 @@ public function show(YotiClient $client)
...
                 ]
             ])
             ->build();
 
         $dynamicScenario = (new DynamicScenarioBuilder())
-            ->withCallbackEndpoint("/account/connect")
+            ->withCallbackEndpoint("/profile")
             ->withPolicy($dynamicPolicy)
             ->withSubject((object)[
                 'subject_id' => "some_subject_id_string"
```
BenSmithYoti, line 28: Rodion, I didn't do the equivalent part for dotnet (DBSCheck example), but I do see that our current version uses "/account/connect" here as the callbackEndpoint  

https://github.com/getyoti/yoti-dotnet-sdk/blob/1ad957543a0f305a3ca5997c32838daa2e45d2da/src/Examples/Profile/CoreExample/Controllers/HomeController.cs#L113

I do also notice an earlier change from a yoti.com to a staging url, so maybe these are related (I've not done much on the Profile/App side - more IDV). 

I don't know what is correct here. My guess is that maybe the dotnet version uses the right endpoint for the staging url and your version works with the production url. 

But just in case it helps you:

https://github.com/getyoti/yoti-dotnet-sdk/commit/166f9af92ba847033a52d7fba18f11faed03d9b0#diff-0f15fcbd73459a6e3a2921f312e0834a0d3bd47a39f2fac0332ac02ab4d76ae7 

Otherwise all LGTM. Thanks

BenSmithYoti (commented):
As per my comment - if you are happy the endpoints are correct please just let me know and I will approve. Looks good, thanks.

examples/profile/app/Http/Controllers/DbsCheckController.php
BenSmithYoti: I've added vince to the reviewers as think java side of this is done, and Ed is on holiday now. Thanks

irotech: about the callback endpoint in the example it should be whatever is configured in the server

.NET has `/account/connect` because of the [endpoint](https://github.com/getyoti/yoti-dotnet-sdk/blob/1ad957543a0f305a3ca5997c32838daa2e45d2da/src/Examples/Profile/CoreExample/Controllers/AccountController.cs#L31)

PHP exposes this [endpoint](https://github.com/getyoti/yoti-php-sdk/blob/e87598934bd6283c9060c45ae4ee1efd33fb9f69/examples/profile/app/Http/Controllers/ProfileController.php#L18)

LGTM at first glance

src/Profile/Service.php
```diff
@@ -88,7 +88,7 @@ public function getActivityDetails(string $encryptedConnectToken): ActivityDetai
 
         // Check response was successful
         if ($receipt->getSharingOutcome() !== self::OUTCOME_SUCCESS) {
-            throw new ActivityDetailsException('Outcome was unsuccessful', $response);
+            throw new ActivityDetailsException($receipt->generateErrorMessage(), $response, $receipt);
```
laurent-yoti, line 91: Seems a bit strange to me to pass a `message` and the `receipt` considering the `message` is actually the result of calling a method on the `receipt`... which then could be done within the `ActivityDetailsException` as it is already doing when calling `receipt->getErrorDetails()`.

Anyway, my 2 cents.

laurent-yoti (approved):
Looks good, interesting to see that we all have slightly different implementations.

PR #292, opened 2022-08-30
SDK-2169

LuborRod:


mehmet-yoti (approved):
Checked issue, compared with other sdk's, seems ok.

PR #294, opened 2022-09-02
Change key in attempts configuration

LuborRod:


laurent-yoti (approved):
Seems logical and correct, but I don't speak PHP ;-)

PR #319, opened 2023-03-06
SDK-2241 Share V2 Create Session

LuborRod:


mehmet-yoti (approved):
After merge we need to change method names to align with other sdks as we agreed

PR #320, opened 2023-03-07
SDK-2250 Create Qr Code

LuborRod:


mehmet-yoti (approved):
After merge we need to change method names to align with other sdks

PR #355, opened 2024-06-20
Sdk 2357 php failure receipt error details

mehmet-yoti:


src/Identity/RequirementNotMetDetails.php
saurabh-yoti, line 1: Some of the fields might not always be present in the $data array. Perhaps better to set these initially as `null`.

saurabh-yoti (approved):

PR #381, opened 2025-06-05
Php 8.4 dev

mehmet-yoti:


examples/digitalidentity/.env.example
```diff
@@ -0,0 +1,14 @@
+0# This file is a template for defining the environment variables
```
saurabh-yoti, line 1: Please remove the leading `0`.

PR #404, opened 2026-03-13
SDK-2793-php-task-response-recommandation added

mehmet-yoti:
### Problem
The `TaskResponse` class did not deserialize or preserve the `recommendation` field returned by the API. When a task response included a recommendation (e.g. `MUST_TRY_AGAIN` with a reason), this data was silently discarded

- **`TaskRecommendationResponse`** — holds `value` (string) and `reason` (TaskRecommendationReasonResponse)
- **`TaskRecommendationReasonResponse`** — holds `value` (string) and `detail` (string)

### Example payload now supported
```json
{
  "type": "ID_DOCUMENT_TEXT_DATA_EXTRACTION",
  "id": "b54287a1-...",
  "state": "DONE",
  "recommendation": {
    "value": "MUST_TRY_AGAIN",
    "reason": {
      "value": "USER_ERROR",
      "detail": "NO_DOCUMENT"
    }
  }
}

### DIFF
--- a/src/DocScan/Session/Retrieve/TaskRecommendationReasonResponse.php
+++ b/src/DocScan/Session/Retrieve/TaskRecommendationReasonResponse.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yoti\DocScan\Session\Retrieve;
+
+class TaskRecommendationReasonResponse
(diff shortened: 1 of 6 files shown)


nikhilPank (approved):
LGTM

PR #419, opened 2026-06-29
Release/4.5.2

### LINKED ISSUE #417: SignedRequestStrategy timestamp may use scientific notation on PHP 8.4, causing signing failures
Hi Yoti team,

We encountered a signing failure when using the PHP SDK on WP Engine FPM running PHP 8.4.22.

The issue appears to come from the timestamp generated in:

```php
Yoti\Http\AuthStrategy\SignedRequestStrategy::createQueryParams()
```

The current upstream code casts the rounded millisecond timestamp to a string:

```php
'timestamp' => (string)(round(microtime(true) * 1000)),
```

In our environment, this can produce a timestamp in scientific notation, for example:

```text
timestamp=1.78231576830E%2B12
```

That timestamp is included in the signed request URL, and Yoti rejects the request with:

```text
401 - MESSAGE_SIGNING - The message was not signed correctly
```

We patched the SDK locally by formatting the timestamp as a plain integer string:

```php
'timestamp' => sprintf('%.0F', round(microtime(true) * 1000)),
```

This produces the expected millisecond timestamp format, for example:

```text
1782315979361
```

After applying that change, the signed request timestamp is emitted as a plain integer and the signing failure is resolved.

Would you consider updating `SignedRequestStrategy::createQueryParams()` to force a non-scientific integer timestamp string?

--

The above was written up from an issue ticket by AI, but I reviewed and debugged and tested the issue (as a Human!). If any more information or testing can be done on the environment in question, please let me know. Happy to help.

mehmet-yoti:
fix: prevent scientific notation in timestamp on low-precision PHP environments (#418)

On PHP FPM environments with non-default precision ini settings (≤12), casting a large float to string could produce scientific notation (e.g. 1.78E+12) instead of a plain integer, causing Yoti signature verification to fail with 401 MESSAGE_SIGNING.

Replace (string)(round(microtime(true) * 1000)) with sprintf('%.0F', microtime(true) * 1000) which forces a plain decimal string regardless of the precision ini setting, and drops the redundant round() since %.0F already rounds to zero decimal places.

Tests added to verify plain integer output under precision=12 and precision=8 (via @dataProvider), and to assert the timestamp falls within the correct Unix-millisecond range under low precision without masking scientific notation via an early (int) cast.

Fixes #417

### DIFF
--- a/src/Http/AuthStrategy/SignedRequestStrategy.php
+++ b/src/Http/AuthStrategy/SignedRequestStrategy.php
@@ -61,7 +61,7 @@ class SignedRequestStrategy implements AuthStrategyInterface
     {
         $params = [
             'nonce' => self::generateNonce(),
-            'timestamp' => (string)(round(microtime(true) * 1000)),
+            'timestamp' => sprintf('%.0F', microtime(true) * 1000),
         ];
 
         if ($this->sdkId !== null) {
--- a/tests/Http/AuthStrategy/SignedRequestStrategyTest.php
+++ b/tests/Http/AuthStrategy/SignedRequestStrategyTest.php
@@ -77,17 +77,19 @@ class SignedRequestStrategyTest extends TestCase
      * can cause (string)(round(microtime(true) * 1000)) to emit scientific notation
      * (e.g. "1.78231576830E+12"), which breaks Yoti signature verification with 401.
      */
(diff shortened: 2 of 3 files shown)


nikhilPank (approved):
LGTM

PR #420, opened 2026-07-02
SDK-2764: PHP - Support handled_check_limit property for checks in IDV sandbox - php

mehmet-yoti:
## Summary

Adds support for the optional \`handled_check_limit\` integer property across all IDV sandbox check-report configuration objects in the PHP SDK sandbox package. The property is omitted from JSON serialization when not set (no-limit behaviour preserved), and only emitted as \`handled_check_limit\` when explicitly configured via the builder. No client-side validation is performed — the backend validates the value.

The companion sandbox-package changes are tracked in [getyoti/yoti-php-sdk-sandbox#71](https://github.com/getyoti/yoti-php-sdk-sandbox/pull/71).

## Changes

- **`SandboxCheck` (base class)** — added nullable `int $handledCheckLimit` property and omit-on-null JSON serialization
- **`SandboxCheckBuilder`** — added `withHandledCheckLimit(int $limit): self` method inherited by all concrete builders
- **Intermediate classes** (`SandboxDocumentCheck`, `SandboxLivenessCheck`, `SandboxIdDocumentComparisonCheck`, `SandboxZoomLivenessCheck`) — forward the new parameter to their parent constructor
- **7 concrete builders** — each passes `$this->handledCheckLimit` when constructing check instances:
  - `SandboxDocumentAuthenticityCheckBuilder`
  - `SandboxDocumentFaceMatchCheckBuilder`
  - `SandboxDocumentTextDataCheckBuilder`
  - `SandboxSupplementaryDocumentTextDataCheckBuilder`
  - `SandboxThirdPartyIdentityCheckBuilder`
  - `SandboxZoomLivenessCheckBuilder`
  - `SandboxIdDocumentComparisonCheckBuilder`
- **7 builder test files** — two new test cases per builder (`shouldBuildWithHandledCheckLimit` and `shouldOmitHandledCheckLimitWhenNotSet`); 14 new tests in total

## QA Test Steps

1. **Setup** — clone the repo, run `composer install`, confirm all existing tests pass with `vendor/bin/phpunit`
2. **Happy path** — instantiate any sandbox check builder (e.g. `SandboxDocumentAuthenticityCheckBuilder`), call `->withHandledCheckLimit(3)->build()`, JSON-encode the result, and assert `"handled_check_limit":3` is present in the output
3. **Omission path** — build the same check without calling `withHandledCheckLimit`, JSON-encode, and assert the key `handled_check_limit` is **absent**
4. **All 7 check types** — repeat steps 2–3 for each of the seven concrete builder classes listed above
5. **Regression** — run the full suite (`vendor/bin/phpunit`) and confirm no pre-existing tests are broken
6. **Edge cases** — pass `withHandledCheckLimit(0)` and `withHandledCheckLimit(1)` and verify the value is serialized correctly; pass a negative int and confirm it serialises without error (backend validation gates rejection)

## Notes

- The property is optional (`nullable int`, defaults to `null`); existing integrations are fully backward-compatible and require no changes
- No client-side range validation is intentional — backend owns the business rule for what constitutes a valid limit
- Follow-up: once the sandbox package PR is merged and released, update the main SDK's `composer.json` dev-dependency on `getyoti/yoti-php-sdk-sandbox` if a version pin needs bumping

---

*Related Jira:* SDK-2764
*Auto-generated by Claude dynamic workflow*


mehmet-yoti:
## Code Review Findings

> Auto-generated by Claude Code (high-effort review). The main SDK PR only modifies `.php-cs-fixer.cache`; all feature changes live in the companion sandbox PR [getyoti/yoti-php-sdk-sandbox#71](https://github.com/getyoti/yoti-php-sdk-sandbox/pull/71). This review covers both.

---

### Critical
None.

### Major
None.

### Minor

**1. `withHandledCheckLimit` returns `self` instead of `static` — breaks fluent-chain type inference**
`SandboxCheckBuilder.php` — `withHandledCheckLimit(int): self`

When a caller chains `->withHandledCheckLimit()` before a subclass-specific method such as `->withDocumentFilter()`, static-analysis tools (PHPStan, Psalm) see the return type as the abstract parent `SandboxCheckBuilder`, which does not expose `withDocumentFilter`, and report a type error. At runtime everything works, but it silently degrades IDE autocomplete and CI static-analysis checks for any integration written against strict levels.

The root cause is pre-existing (all other builder methods also return `self`), but adding a new entry point makes the pattern more visible and compounds the problem.

**Suggested fix:**
```php
public function withHandledCheckLimit(int $handledCheckLimit): static
{
    $this->handledCheckLimit = $handledCheckLimit;
    return $this;
}
```
(PHP 8.0+ supports `static` as a return type; the companion sandbox CI should confirm the minimum PHP version supports it.)

---

**2. No test for `withHandledCheckLimit(0)` edge case**
All seven builder test files — `shouldBuildWithHandledCheckLimit` tests

The PR QA steps explicitly list `withHandledCheckLimit(0)` as an edge case requiring verification. No unit test covers it. While PHP's `isset(0)` returns `true` (so `0` is serialised correctly), the absence of a test means a future change to the null-guard (e.g. switching from `isset` to a falsy check such as `if ($this->handledCheckLimit)`) would silently break the `0` case without failing any test.

**Suggested fix:** Add a `shouldBuildWithHandledCheckLimit_zero` test in at least one builder test class (e.g. `SandboxDocumentAuthenticityCheckBuilderTest`):
```php
public function shouldBuildWithHandledCheckLimitZero(): void
{
    $result = (new SandboxDocumentAuthenticityCheckBuilder())
        ->withRecommendation($this->recommendationMock)
        ->withHandledCheckLimit(0)
        ->build();

    $this->assertStringContainsString('"handled_check_limit":0', json_encode($result));
}
```

---

### Nit

**3. `isset()` used instead of `!== null` in `SandboxCheck::jsonSerialize()`**
`src/DocScan/Request/Check/SandboxCheck.php` line ~35

```php
if (isset($this->handledCheckLimit)) {
```

`isset()` returns `false` for `null` and for unset variables; `!== null` explicitly tests for `null`. For a property that is always initialised in the constructor (and is declared `?int`), both are equivalent, but `!== null` more directly expresses the design intent ("only emit when a value was explicitly set") and is consistent with how nullable properties are guarded elsewhere in the SDK (e.g. `SandboxIdDocumentComparisonCheck::jsonSerialize` already uses `isset` for the filter — so the nit applies consistently, but it is worth noting here as the property is integer-typed, where falsy confusion is more likely).

**Suggested fix:**
```php
if ($this->handledCheckLimit !== null) {
    $data->handled_check_limit = $this->handledCheckLimit;
}
```

---

**4. Main SDK PR (#420) contains no PHP source changes**
`.php-cs-fixer.cache`

The commit is titled `feat(SDK-2764): PHP - Support handled_check_limit property for checks in IDV sandbox [php]`, but the only file changed in the main `yoti-php-sdk` repository is the CS-fixer cache. The entire feature implementation is in [getyoti/yoti-php-sdk-sandbox#71](https://github.com/getyoti/yoti-php-sdk-sandbox/pull/71). This creates a disconnect: a reviewer checking only PR #420 will find no code to review, and the git history for the main SDK will contain a misleading feat-commit with no accompanying source delta.

Consider either:
- Squashing the cache change into a `chore:` commit referencing the sandbox PR, or
- Keeping the cache update as a separate `chore: update .php-cs-fixer.cache` commit and removing the `feat` prefix until the sandbox package version is pinned in `composer.json`.

